PR #2, opened 2022-05-29
operator

jmazzitelli:
This introduces the OSSM Plugin operator.

Note that I moved the plugin make targets to make/Makefile.plugin.mk, but they all work as before. So nothing changes from the user-perspective (e.g. `make build-plugin build-plugin-image` still works as does `make deploy-plugin enable-plugin`). I also added a `make undeploy-plugin`.

Do a `make help` to see all the make targets available.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,23 +1,49 @@
-# NOTE: This makefile is opinionated and it uses "oc" (instead of kubectl) and "podman" (instead of docker)
+SHELL=/bin/bash
 
-clean-plugin:
-	@rm -rf ${ROOTDIR}/plugin/node_modules
-	@rm -rf ${ROOTDIR}/plugin/dist
+# Identifies the current build.
+VERSION ?= v0.0.1-SNAPSHOT
+COMMIT_HASH ?= $(shell git rev-parse HEAD)
 
-build-plugin: clean-plugin
-	cd plugin && yarn install && yarn build
+# Directories based on the root project directory
+ROOTDIR=$(CURDIR)
+PLUGIN_DIR=${ROOTDIR}/plugin
+OPERATOR_DIR=${ROOTDIR}/operator
 
-build-plugin-image:
-	cd plugin && podman build -t quay.io/kiali/servicemesh-plugin:latest .
+# Determine if we should use Docker OR Podman - value must be one of "docker" or "podman"
+DORP ?= podman
 
-build-plugin-push:
-	cd plugin && podman push quay.io/kiali/servicemesh-plugin:latest
+# Find the oc client executable
+OC ?= $(shell which oc 2>/dev/null || echo "MISSING-OC-FROM-PATH")
 
-deploy-plugin:
-	cd plugin && oc apply -f manifest.yaml
+# list for multi-arch image publishing
+TARGET_ARCHS ?= amd64 arm64 s390x ppc64le
 
-enable-plugin:
-	oc patch consoles.operator.openshift.io cluster --patch '{ "spec": { "plugins": ["servicemesh"] } }' --type=merge
+# Local arch details needed when downloading tools
+OS := $(shell uname -s | tr '[:upper:]' '[:lower:]')
+ARCH := $(shell uname -m | sed 's/x86_64/amd64/')
 
-restart-plugin:
-	oc rollout restart deployments/servicemesh-plugin -n servicemesh-plugin
+include make/Makefile.plugin.mk
+include make/Makefile.operator.mk
+include make/Makefile.cluster.mk
+include make/Makefile.olm.mk
+
+help:
+	@echo
+	@echo "Plugin targets - used to develop and build the plugin"
+	@sed -n 's/^##//p' make/Makefile.plugin.mk | column -t -s ':' |  sed -e 's/^/ /'
+	@echo
+	@echo "Operator targets - used to develop and build the operator"
+	@sed -n 's/^##//p' make/Makefile.operator.mk | column -t -s ':' |  sed -e 's/^/ /'
+	@echo
+	@echo "Cluster targets - used to manage images on the remote cluster"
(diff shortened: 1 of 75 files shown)


jmazzitelli:
This is a WIP. The "make run-playbooks" works - it will deploy (and then undeploy) the OSSM Plugin using the operator playbooks. So I know that works.

There is still a lot left to do. I need to actually deploy the operator in the cluster and see it work. I need make targets to push/deploy the operator. Need to complete simple molecule tests just to confirm the basics can be tested. Also need to see this deploy using the OLM metadata. The metadata validates, just need to bundle it up and test it (hopefully I can get make targets to do this also). So, again, lots to do but the basic functionality in the ansible scripts appears to work.

jmazzitelli:
Getting closer :)
image

jmazzitelli:
Operator is now being deployed successfully via make targets.

image

jmazzitelli:
Operator is now able to install the plugin - I see this in the side menu of the console:

image

jmazzitelli:
I confirmed uninstalling the plugin works also - just delete the OSSMPlugin CR and the plugin is deleted and disabled from the console.

Just a couple more things to do before this PR can get ready for review and merge. I would like the get at least one basic molecule test written and working. Also, last time I checked the `run-operator` make target wasn't working (this allows you to run the operator on the local box without having to deploy it in the cluster - similar to `run-playbook` (which does work) only `run-operator` runs what the actual operator pod would be running (with the ansible operator base image handling the ansible playbook invocations).

Some other little odds-and-ends - I want to get the make infrastructure to auto-discovery the kiali URL so it can populate it inside the test CR when `make install-cr` is run - make it easy on the devs. Also want to confirm the generation of the CRD docs still works and validation of the CR still works (both did work last I tried which was yesterday)

jmazzitelli:
This is how you can test the basic functionality of the OSSM Plugin operator.

<tl;dr>Install Istio and Kiali and then run `make operator-create install-cr`

1. You need to first install Istio and Kiali. The make infrastructure will attempt to auto-discover the Kiali route URL so it can tell the OSSM plugin where the Kiali endpoint is.
2. Install the operator via OLM: `make operator-create`. It will take OLM a few seconds to deploy the catalog source and start up the operator.
3. Install the latest OSSM plugin published on quay.io: `make install-cr`. It will take a few seconds but the OpenShift Console will tell you when it has been updated - at that point, you can refresh the console and see the "Service Mesh" options in the  Console.
4. To undeploy the plugin but keep the operator running, run: `make uninstall-cr`
5. To uninstall everything (the CR and operator), run: `make operator-delete`

lucasponce:
Thanks, @jmazzitelli, I'm going to start reviewing it.

I have one early comment about operator naming in the way that is seen by users:

image

Instead, to use the "OSSM" acronym in the title why not use "Service Mesh Plugin" or "Kiali Service Mesh Plugin"? 

I guess "OSSM" is mostly used internally and perhaps something that describes the plugin in the "OperatorHub" can help to be seen when users play with "service mesh" tech.

lucasponce:
I get the "Failed to pull image" after following the steps:

jmazzitelli:
Yes, naming was one thing I wanted to have a discussion on.

I see you used "Service Mesh" for the plugin (e.g. "servicemesh-plugin" image name and the name of the plugin is "servicemesh" in the JS code). But the product uses the term "OpenShift Service Mesh", and even notice this github repo is named "openshift-servicemesh-plugin".

To me, the term "Service Mesh" is generic - everyone has a "service mesh" :). The name "OpenShift Service Mesh" and its abbreviation "OSSM" specifies this specific project/product (the upstream is Maistra which theoretically could run on non-OpenShift clusters; however, the downstream is called "OpenShift Service Mesh"). As you see:
image


So, I used with the names "OpenShift Service Mesh", "OSSM Plugin" and "OSSM" to retain consistency across the projects. I don't think we should be using just the word "Service Mesh" or "servicemesh" - that is a "general" term that does not identify this specific project/product and has no direct relationship with Kiali when compared to "OpenShift Service Mesh".

jmazzitelli:
> I get the "Failed to pull image" after following the steps:

Opps. I forgot one thing in the instructions (I updated the comment above). You have to build and push the operator image to the internal image registry. So log into the internal image registry using "podman login" and then run `make cluster-push-operator`.

If you don't know the internal image registry hostname, just run `make cluster-status` and look at the bottom of its output - it will tell you want login command to run. For example:

```
$ make cluster-status

.... a bunch of output here...
=================================================================
Image Registry login: podman login --tls-verify=false -u kiali -p $(/bin/oc whoami -t) default-route-openshift-image-registry.apps.my-cluster.com

...

jmazzitelli:
> note that the plugin mostly provides a "Service Mesh" logical concept in navigation and tabs.

I think it's fine for the Console to show "Service Mesh" in the tabs - since "OpenShift Service Mesh" would be too long in the UI to display.

For docs and for component names (like deployments and things like that), I think using the full name of "openshift-servicemesh" (or "ossm" for short) should be used. It makes it clear these docs and components are part of the OSSM offering.

lucasponce:
All looks good:

image
image
image

Note that the "Service Mesh" tab under the project is only available in the "latest" OpenShift Console, but that's out of the scope of the plugins; it can be seen in the image used for local development.

lucasponce (approved):
It's still in draft, but it looks ok to me; I've tested the instructions provided.

lucasponce:
Side comment, perhaps the README.md in the root folder should be updated, but to me it's fine to consolidate and iterate there on these first steps.

jmazzitelli:
darn it! I went to click "Command" and hit the "Close with Comment" button instead!.. reopening...

lucasponce:
>  We can iterate on it.

+1

jshaughn:
@jmazzitelli thanks for this so quickly!   As for naming, I'm +1 for OpenShift but -1 for OSSM.

PR #3, opened 2022-05-31
operator will auto-discover kiali route if kiali.url not specified in CR

jmazzitelli:
You can now leave the `spec.kiali.url` unspecified (or set it to empty string `""`, which is the default) in the CR, and the operator will attempt to auto-discover it. It will first give precedence to any Kiali route it finds in the same namespace where the CR is, otherwise, it will look for any Kiali installed anywhere in the cluster and use that.

You can always override this auto-discovery mechanism by setting `spec.kiali.url` in the CR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ include make/Makefile.plugin.mk
 include make/Makefile.operator.mk
 include make/Makefile.cluster.mk
 include make/Makefile.olm.mk
+include make/Makefile.molecule.mk
 
 help:
 	@echo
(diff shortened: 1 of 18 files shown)


jmazzitelli:
I will be adding another commit to this PR - the molecule tests. Just very basic but will test that the route url is picked up.

jmazzitelli:
This test passes - it confirms the auto-discovered url makes its way to the plugin configmap:

`make -e MOLECULE_SCENARIO="config-values-test" cluster-push-operator operator-create molecule-test`

lucasponce (approved):
LGTM

PR #7, opened 2022-06-02
Propagate session token

lucasponce:
This work propagates the OpenShift Console session token into Kiali using the new change introduced in https://github.com/kiali/kiali/pull/5133

When the OpenShift console talks with the plugins, it already propagates the session token, with a change in the configuration, the Nginx server can add that info in the response header.

At the same time that the plugin asks for the Kiali URL, the token is fetched and propagated to Kiali.

How to test:

(1) Validate that the user hasn't logged into Kiali:
image

(2) Access the OpenShift Console, and validate that the user is logged into Kiali WITHOUT requiring the previous login in the standalone app:
image

(3) Check that being logged into OpenShift Console DOESN'T log the user into Kiali standalone, that authentication only should work in the context of the OpenShift Console (or a special bookmark containing the credentials).

### DIFF
--- a/operator/roles/default/ossmplugin-deploy/templates/openshift/configmap-nginx.yaml
+++ b/operator/roles/default/ossmplugin-deploy/templates/openshift/configmap-nginx.yaml
@@ -6,7 +6,7 @@ metadata:
   labels: {{ ossmplugin_resource_metadata_labels }}
 data:
   nginx.conf: |
-    error_log /dev/stdout info;
+    error_log /dev/stdout;
     events {}
     http {
       access_log         /dev/stdout;
@@ -17,6 +17,9 @@ data:
         listen              9443 ssl;
         ssl_certificate     /var/serving-cert/tls.crt;
         ssl_certificate_key /var/serving-cert/tls.key;
+
+        add_header oauth_token "$http_Authorization";
+
         location / {
           root                /usr/share/nginx/html;
         }
--- a/plugin/manifest.yaml
+++ b/plugin/manifest.yaml
@@ -74,17 +74,21 @@ metadata:
     app.kubernetes.io/part-of: servicemesh-plugin
 data:
   nginx.conf: |
-    error_log /dev/stdout info;
+    error_log /dev/stdout;
     events {}
     http {
       access_log         /dev/stdout;
       include            /etc/nginx/mime.types;
       default_type       application/octet-stream;
       keepalive_timeout  65;
+
       server {
         listen              9443 ssl;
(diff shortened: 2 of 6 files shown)


lucasponce:
Hey @jmazzitelli, I've tested this locally, but if you can give a try to this PR, just to check that it didn't miss anything, that would be great.

lucasponce:
Hey @jmazzitelli, I've tested this locally, but if you can give a try to this PR, just to check that it didn't miss anything, that would be great.

jmazzitelli:
@lucasponce I can't get this to work.

Built and deployed kiali from master, built the plugin from your PR, deployed everything:

image

Kiali log:
```
2022-06-06T12:30:05Z INF Kiali: Version: v1.52.0-SNAPSHOT, Commit: 8191532af24f5f93c9534eada25d0557bbc0996a
2022-06-06T12:30:05Z INF Using authentication strategy [openshift]
```

So you can see I'm currently running a build whose commit includes the new auth code enhancement (commit 8191532a is master as of right now).

One problem with the plugin is it does not log its version in its output logs - so I cannot confirm what it is running, but I do have your PR branch checked out :

```
$ git branch --show-current
lucasponce-propagate-credentials-on-iframe
```

and installed using the make targets which should have pushed and ran this dev build... and I see the pod is pulling my dev build:

```
$ oc get pods -n ossmplugin -l app.kubernetes.io/name=ossmplugin -o jsonpath={.items[0].spec.containers[0].image}
image-registry.openshift-image-registry.svc:5000/kiali/servicemesh-plugin:dev.
```

So it looks like its running the latest.

lucasponce:
Could you debug the OpenShift Console DOM and check the URL that is passed to the iframe?
(With a browser debugger)

lucasponce:
Another minor test, if you use the "kubeadmin" instead of "kiali"
image

do you have the same result?

(Just brainstorming if the user may need something in terms of permissions)

jmazzitelli:
same with kubeadmin:
image

lucasponce:
Ok, this PR updated the `plugin/manifest.yaml` but not the operator; that's one thing to address on this.

Also, in the @jmazzitelli's environment, the plugin request has the authentication in the cookie instead of the header, then the Nginx server is not adding it to the response header.

I have a different OpenShift Console version, but I think both cases should be supported, I'll update the PR to cover that.

Thanks for the feedback @jmazzitelli

jmazzitelli (approved):
all works.
Login is not required when using the plugin - kiali is already logged in.
When I go to standalone kiali, I still have to log in manually as expected.

PR #12, opened 2022-06-03
add some descriptions to csv metadata

jmazzitelli:


jmazzitelli:
When creating the OSSMPlugin CR from the OpenShift Console UI, you will now get the ability to enter the Kiali UI within a text field that shows some help text:

image

jmazzitelli:
and here is what the operator page looks like in the OpenShift Console:

image

jmazzitelli:
This is a good beginning point. We can iterate later on what we want these data to show and if we want to add more.

We could do the following in the future:
1. Add more description text to explain further what this is
2. Add additional links (to docs? to other source code repos like kiali/kiali)
3. Maybe a different icon?

Issue #16, opened 2022-06-06
plugin should log its version (git commit or some build version)


jmazzitelli:
we need to identify the plugin that is running - we have a version label on the pod, but it would be nice to log some version, too. This will be helpful for those running a dev version deployed outside of the operator or even when deployed with the operator but using the make targets to push a dev build.

PR #17, opened 2022-06-06
plugin will log version/commit at startup

jmazzitelli:
fixes: https://github.com/kiali/openshift-servicemesh-plugin/issues/16

to test, just run the plugin in the cluster and look at the pod logs - the first several lines will look like this - see the version/commit that prints out:

```
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/01-log-version.sh
OpenShift Service Mesh Plugin: Version=[v0.0.1-SNAPSHOT], Commit=[e699245fa9953e6678e009da17fae2fd61e05c7e]
```

This PR only adds that one script which prints out that one line to the output: `OpenShift Service Mesh Plugin: Version=[v0.0.1-SNAPSHOT], Commit=[e699245fa9953e6678e009da17fae2fd61e05c7e]`

### DIFF
--- a/make/Makefile.plugin.mk
+++ b/make/Makefile.plugin.mk
@@ -24,7 +24,7 @@ build-plugin:
 
 ## build-plugin-image: Builds the plugin and its container image.
 build-plugin-image: build-plugin
-	${DORP} build -t ${PLUGIN_QUAY_TAG} ${PLUGIN_DIR}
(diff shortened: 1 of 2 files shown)


lucasponce (approved):
image

PR #19, opened 2022-06-06
add info on how to release the operator

jmazzitelli:


jmazzitelli:
We can't test these instructions yet because we don't want to release :) But I'm pretty sure these will work. We will iterate on these instructions and correct them as necessary when we go to release for the first time, but I wanted to get this in now so we are prepared.

PR #20, opened 2022-06-06
put kiali version in CR status so we know what the plugin is talking to

jmazzitelli:
Add new `kialiVersion` field in the CR `status`. This is going to be useful when people are installing the plugin and things aren't working - we will need to know what version of Kiali the plugin is talking to.

This is what the status field will now look like (notice the new kialiVersion field):

```sh
$ oc get -n ossmplugin ossmplugins.kiali.io ossmplugin -o jsonpath={.status.environment} | jq
```

```json
{
  "kialiVersion": "v1.52.0-SNAPSHOT",
  "kubernetesVersion": "1.23.3+e419edf",
  "openshiftVersion": "4.10.5",
  "operatorVersion": "v0.0.1"
}

```

### DIFF
--- a/operator/roles/default/ossmplugin-deploy/tasks/main.yml
+++ b/operator/roles/default/ossmplugin-deploy/tasks/main.yml
@@ -107,13 +107,20 @@
   when:
   - ossmplugin_vars.kiali.url == ""
 
+- name: Determine Kiali version
+  set_fact:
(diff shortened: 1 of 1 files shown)


lucasponce:
I've tested this but I saw it's moved to draft, just in case if it helps:
image

PR #23, opened 2022-06-14
Fix security issues in plugin Dockerfile

aljesusg:
Waiting for security results in my local repo https://quay.io/repository/aljesusg/kiali-plugin?tab=tags

I don't know if there is any way to check this locally


plugin/Dockerfile
```diff
@@ -1,16 +1,36 @@
...
+USER 1001
 
-RUN chmod g+rwx /var/cache/nginx /var/run /var/log/nginx
-COPY --from=build /usr/src/app/dist /usr/share/nginx/html
 
 # Have the plugin log its version immediately upon startup
 ARG VERSION
 ARG COMMIT_HASH
-RUN echo "echo 'OpenShift Service Mesh Plugin: Version=[${VERSION}], Commit=[${COMMIT_HASH}]'" > /docker-entrypoint.d/01-log-version.sh
-RUN chmod ug+x /docker-entrypoint.d/01-log-version.sh
```
jmazzitelli, line 16: This breaks the feature we just added last week: https://github.com/kiali/openshift-servicemesh-plugin/pull/17

jmazzitelli (changes requested):
This needs to somehow provide a way to log the version/hash like we did before this change.

plugin/Dockerfile
aljesusg: Yeah i am checking how to do it but there is not dockerentry in the iamge registry.access.redhat.com/ubi8/nginx-120 taht's why I set this like a draft...researching

aljesusg: Let me try to have this in log. If not I don't see bad idea to have an endpoint for this

aljesusg:
I fixed the message @jmazzitelli 

Security issues reduced from 15 critical 146 fixable to 4 high 
The issue reported is related with CVE-2020-26137 urllib3 there is not fix yet

On the other hand the main one was the xz libs reported by our workmate ...so I hope to be able to update this when they publish the last ubi8 image witht his fix (So we'll remove the line 20)

# Demo
Screenshot from 2022-06-15 18-20-05

plugin/Dockerfile
```diff
@@ -1,16 +1,38 @@
...
\ No newline at end of file
+RUN printf "\necho 'OpenShift Service Mesh Plugin: Version=[${VERSION}], Commit=[${COMMIT_HASH}]' >> /proc/1/fd/1" >> ${NGINX_CONTAINER_SCRIPTS_PATH}/common.sh
+
+USER 1001
+LABEL com.redhat.component="openshift-service-mesh-plugin" \
+      name="openshift-service-mesh-plugin" \
+      summary="OpenShift Service Mesh Kiali Plugin" \
```
lucasponce, line 32: "OpenShift Service Mesh Kiali Plugin" -> "OpenShift Service Mesh Plugin"

plugin/Dockerfile
```diff
@@ -1,16 +1,38 @@
...
+      description="Kiali plugin for Openshift Service Mesh" \
```
lucasponce, line 33: "Kiali plugin for Openshift Service Mesh" -> "The OpenShift Service Mesh Plugin adds Kiali functionality to the OpenShift Console."

Wording for this effort should be synced.

The operator would have the "master" info about it.

jmazzitelli:
This causes the build to be very slow (its been over 5 minutes and I'm still not done). So I created this issue: https://github.com/kiali/openshift-servicemesh-plugin/issues/25

jmazzitelli:
I am unable to build - it might be a infrastructure issue, but it hangs here:

```
[2/2] STEP 3/11: RUN yum -y upgrade xz-libs
Updating Subscription Management repositories.
Unable to read consumer identity
Subscription Manager is operating in container mode.
Red Hat Enterprise Linux 8 for x86_64 - AppStre 2.8 MB/s |  44 MB     00:15    
Extra Packages for Enterprise Linux 8 - x86_64  308 kB/s |  11 MB     00:37    
Extra Packages for Enterprise Linux Modular 8 -  15 kB/s | 1.0 MB     01:08    
```

jmazzitelli (changes requested):
I would like to block this from being merged until we can figure out why the build is so slow/hung. It still isn't building for me, and it is now the second time I tried.

Or, incorporate this into the PR so people who need to build the plugin can circumvent this by building a dev container: https://github.com/kiali/openshift-servicemesh-plugin/issues/25

plugin/Dockerfile
```diff
@@ -1,16 +1,38 @@
...
+
+# Add application sources to a directory that the assemble script expects them
+# and set permissions so that the container runs without root access
+USER 0
+
+# Install upgrades security
+#   xz-libs RHSA-2022:4991
+#   urllib3 CVE-2020-26137 CVE-2021-33503  Not available in the ubi8
+#
+RUN yum -y upgrade xz-libs
```
jmazzitelli, line 20: This is causing my build to hang. See https://github.com/kiali/openshift-servicemesh-plugin/issues/25 for my suggestions - don't do this yum upgrade if env var DEV_MODE is set to "true".

jmazzitelli:
Just ran this... I see:

```
OpenShift Service Mesh Plugin: Version=[0], Commit=[bbd0136c86ef7b5cc2f4f8cfba131ac99d743e27]
```

The version is wrong here. Should be the VERSION env var.

plugin/Dockerfile
aljesusg: > Just ran this... I see:
> 
> ```
> OpenShift Service Mesh Plugin: Version=[0], Commit=[bbd0136c86ef7b5cc2f4f8cfba131ac99d743e27]
> ```
> 
> The version is wrong here. Should be the VERSION env var.
> 
> I can't explain why the version is showing as "0" - it all looks correct in the dockerfile/makefile. I'm looking into it. It doesn't make sense why this doesn't work.

mmm let me check it.

About slowly....is not my case in my env the dockerfile was build so fast....anyway the "yum" instruction is a temporal solution, they are going to update the image soon

aljesusg:
> Just ran this... I see:
> 
> ```
> OpenShift Service Mesh Plugin: Version=[0], Commit=[bbd0136c86ef7b5cc2f4f8cfba131ac99d743e27]
> ```
> 
> The version is wrong here. Should be the VERSION env var.
> 
> I can't explain why the version is showing as "0" - it all looks correct in the dockerfile/makefile. I'm looking into it. It doesn't make sense why this doesn't work.
> 
> UPDATE: I think I see the problem. There is already a VERSION env var set - and according to Dockefile rules, ENV overrides ARG. So we need to pass in ARG as a different name - PLUGIN_VERSION.
> 
> This works (I prefix COMMIT_HASH with PLUGIN_ for consistency):
> 
> make/Makefile.plugin.mk:
> 
> ```diff
> -       ${DORP} build --build-arg VERSION="${VERSION}" --build-arg COMMIT_HASH="${COMMIT_HASH}" -t ${PLUGIN_QUAY_TAG} ${PLUGIN_DIR}
> +       ${DORP} build --build-arg PLUGIN_VERSION="${VERSION}" --build-arg PLUGIN_COMMIT_HASH="${COMMIT_HASH}" -t ${PLUGIN_QUAY_TAG} ${PLUGIN_DIR}
> ```
> 
> plugin/Dockerfile:
> 
> ```diff
> - ARG VERSION
> - ARG COMMIT_HASH
> - RUN echo "echo 'OpenShift Service Mesh Plugin: Version=[${VERSION}], Commit=[${COMMIT_HASH}]'" > /docker-entrypoint.d/01-log-version.sh
> + ARG PLUGIN_VERSION
> + ARG PLUGIN_COMMIT_HASH
> + RUN echo "echo 'OpenShift Service Mesh Plugin: Version=[${PLUGIN_VERSION}], Commit=[${PLUGIN_COMMIT_HASH}]'" > /docker-entrypoint.d/01-log-version.sh
> ```
> 
> and
> 
> ```diff
> -       version=$VERSION \
> -       commit=$COMMIT_HASH
> +       version=$PLUGIN_VERSION \
> +       commit=$PLUGIN_COMMIT_HASH
> ```

I renamed to VERSION_PLUGIN and works....maybe this image is overwrite the VERSION variable or something like this

jmazzitelli:
> I renamed to VERSION_PLUGIN and works....maybe this image is overwrite the VERSION variable or something like this

It's because there is a ENV "VERSION" already defined, probably in the parent base image. Just rename to PLUGIN_VERSION (and I used PLUGIN_COMMIT_HASH, too, for consistency) and all works.

plugin/Dockerfile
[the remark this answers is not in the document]
jmazzitelli: The problem has gone away. I can build now. So ignore this DEV_MODE thing for now. Don't put it in this PR

jmazzitelli (approved):
This works now. If I `make cluster-push` the image builds and pushes. And in the logs I can see the correct verson/hash:

```
$ podman run --rm -i -t quay.io/kiali/servicemesh-plugin:dev
OpenShift Service Mesh Plugin: Version=[v0.0.1-SNAPSHOT], Commit=[abb0089b77dd79c706a11fe706bb754056be1fd0]
```

Issue #21, opened 2022-06-08
Prototype a infrastructure to customize links in Kiali under Plugin environment


lucasponce:
Kiali has multiple locations with links to other features, typically Graph -> Details, Overview -> Lists, etc.

Under these first steps of this plugin effort, one of the added values of this integration would be to provide links to connect the embedded pages of the Plugin with other features of the OpenShift Console and/or other features inside Kiali.

The final stories of these features are not yet defined but I think we have enough information to prototype a mechanism that provides Kiali with all necessary to perform some conditional rendering of the links, and under the kiosk/plugin mode it can add a customized link.

The goal of this issue is just to place some initial functionality in both (plugin and Kiali) to allow customized links under the kiosk mode.

Converting this into a task to cover the following scenarios:
- [x] Overview page navigation to Graph/Istio Config https://github.com/kiali/openshift-servicemesh-plugin/pull/24
- [x] Graph page navigation to Istio Config and Service Mesh tab details https://github.com/kiali/openshift-servicemesh-plugin/pull/27
- [x] Service Mesh tabs navigation to OpenShift Console details https://github.com/kiali/openshift-servicemesh-plugin/pull/41

lucasponce:
The basic navigation defined in the plugin from new pages to the existing ones is completed.

PR #24, opened 2022-06-15
Introduce listeners for plugin navigation

lucasponce:
Related to https://github.com/kiali/openshift-servicemesh-plugin/issues/21

This work integrates the changes from https://github.com/kiali/kiali/pull/5211 into the plugin.

It captures the messages from Kiali to perform the navigation inside the plugin:
image

When a user clicks on the "Overview" actions, the user will navigate to the "Graph" or "Istio Config" page of the plugin:
image

Time and duration parameters are propagated.

There are some changes in the links on embedded mode, that I think may more sense:
image

Status links (healthy/degraded/failures) should be redirected to the graph but with a "find filter" applied:
image

That can simplify some features that OpenShift Console doesn't provide (i.e. a list of applications, or a multi-namespace list).

This work is only scoped to the "Service Mesh -> Overview" page, where all the links should be now handled.

Other pages (i.e Graph, Istio Config, or the Service Mesh tabs) are not yet implemented.

### DIFF
--- a/plugin/src/components/GraphPage.tsx
+++ b/plugin/src/components/GraphPage.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {kioskUrl, properties} from "../properties";
+import {initKialiListeners, kioskUrl, properties} from "../properties";
 import {consoleFetch } from "@openshift-console/dynamic-plugin-sdk";
 
 const GraphPage = () => {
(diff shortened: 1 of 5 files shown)


jshaughn (approved):
The ability to intercept and redirect urls is pretty neat.  Do you think there is a way to redirect to a non-kiali page?

lucasponce:
> Do you think there is a way to redirect to a non-kiali page?

Oh, yes, indeed, that's what it does. The pages of the plugin are part of the navigation of the OpenShift Console, on this case, they navigate inside the plugin pages but it can point to any location.

PR #27, opened 2022-06-17
Graph kiosk links

lucasponce:
Related to https://github.com/kiali/openshift-servicemesh-plugin/issues/21

This PR adds links from Service Mesh -> Graph to OpenShift Console.

When the user operates in the plugin, the graph will link inside the OpenShift console.

What to expect:
- Context menu.
- Sidepanel links (including traces and spans).

Note, that there is a case that is not fully resolved, but I think the first approach is not bad: OpenShift doesn't have a solid "app" entity/concept as it exists on Kiali. So, in these cases, I voted to redirect the user to the specific pods/workloads filtered by the "app" label, as it's the most "correct" map in terms of concepts.

There are other "app" groups in the "Developer Topology" and other potential "soft" groups, but to not open those discussions yet, I think this approach could be a reasonable workaround.


jmazzitelli:
One thing that we will need to address (probably not in the PR - maybe I should write up a separate issue?) is that we need these links to support multi-tenancy.

OSSM can create multiple control planes in a cluster, and Kiali is installed in each of them. We need the plugin to talk to the correct Kiali.

Right now, the plugin relies on the single config setting kiali.url. However, that should only be the fallback URL (in the case of `accessible_namespaces = [**]` for example). But for the actual OSSM use case, we should look at the namespace labels, determine which Kiali is associated with that namespace and then we can get its route from that.

The Kiali operator will label each namespace with a kiali-specific label that identifies the Kiali that accesses that namespace. From that name identifier, we should be able to find the route.

@lucasponce I don't think this is related to this PR - so should I write this up as a separate issue? Or do you want to do this in this PR, too?

lucasponce:
> One thing that we will need to address (probably not in the PR - maybe I should write up a separate issue?) is that we need these links to support multi-tenancy.

Well, more than links to me multi-tenancy is a news story that we need to figure out how to address it. We discussed with Jamie and Andrew that is not bad to assume a single mesh per cluster in these first steps.

Because the plugin would need also to have visibility of the multi-tenant installation as a concept to see how to deal with one or another.

It's a good point and we can open a story for it but I think that when we have a robust use case for single cluster + single mesh, the plugin could be ready to collect users' feedback.

lucasponce:
> so should I write this up as a separate issue? Or do you want to do this in this PR, too?

+1 we can put it in the backlog, no problem on that.

lucasponce:
This PR and this https://github.com/kiali/kiali/pull/5223 are ready to be tested together.

A small demo about the scope of these new capabilities:
plugin-graph-links
r

plugin/src/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+                    detailUrl = '/k8s/ns/' + namespace + '/deployments' + detail.substring('/workloads'.length) + '/servicemesh';
+                    detailUrl += webParamsIndex > -1 ? webParams : '';
+                }
+                if (detail.startsWith('/services')) {
+                    // OpenShift Console has a "services" list page
+                    detailUrl = '/k8s/ns/' + namespace + detail + '/servicemesh';
+                    detailUrl += webParamsIndex > -1 ? webParams : '';
+                }
+
+                console.log('TODELETE detailUrl: ' + detailUrl);
```
leandroberetta, line 79: All this reminders needs to be deleted on this PR?.

leandroberetta (approved):
Tested locally, works good!.

plugin/src/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+                history.push(servicemeshUrl);
+            }
+            if (kialiAction.startsWith('/istio')) {
+                const istioConfigUrl = kialiAction.replace('istio', 'istioconfig');
+                history.push(istioConfigUrl);
+            }
+            if (kialiAction.startsWith('/namespaces')) {
+                const webParamsIndex = kialiAction.indexOf('?');
+                const webParams = webParamsIndex > -1 ? kialiAction.substring(webParamsIndex) : '';
+                console.log('TODELETE webParams: '+ webParams);
```
jmazzitelli, line 49:
```suggestion
```

plugin/src/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+
+                const namespacesLength = '/namespaces/'.length;
+                const namespace = kialiAction.substring(namespacesLength,  kialiAction.indexOf('/', namespacesLength + 1));
+                console.log('TODELETE namespace: '+ namespace);
```
jmazzitelli, line 53:
```suggestion
```

plugin/src/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+
+                const detail = kialiAction.substring(namespacesLength + namespace.length, webParamsIndex > -1 ? webParamsIndex : kialiAction.length);
+                console.log('TODELETE detail: '+ detail);
```
jmazzitelli, line 56:
```suggestion
```

plugin/src/utils.ts
jmazzitelli, line 79:
```suggestion
```

jmazzitelli (changes requested):
need to remove those temporary debug log messages

Aside from that, this all runs good. We can merge this after those "TODELETE" log messages are removed

jmazzitelli (approved):

Issue #34, opened 2022-06-27
error: plugin__servicemesh.json is missing


jmazzitelli:
This is a minor issue but I wanted to document it.

After submitting the very first request to the plugin, this error appears in the pod logs. It doesn't look like it affects the plugin operation (things still work), but since it is flagged as an "error" I wanted to report this.

`2022/06/27 12:54:21 [error] 18#0: *1 open() "/usr/share/nginx/html/locales/en/plugin__servicemesh.json" failed (2: No such file or directory), client: 10.130.0.132, server: , request: "GET /locales/en/plugin__servicemesh.json HTTP/1.1", host: "ossmplugin.ossmplugin.svc.cluster.local:9443", referrer: "https://console-openshift-console.apps...com/k8s/cluster/projects"`

lucasponce:
Yes, thanks for documenting it.
In short, the OpenShift console supports i18n, and at some point that will be something to consider.
But Kiali doesn't support it and it's weird to introduce it in the plugin only.

So, having this issue filled would help to address it in a future iteration as it may have more implications that perhaps today we can't open just yet.

lucasponce:
Well, to solve this issue, the easy thing is to add a file with this English support as default to remove the error.
The i18n is a different epic, I'd say.

aljesusg:
Let me address this. I saw it in another plugins

PR #36, opened 2022-06-28
Fix error about translation

aljesusg:
Fix https://github.com/kiali/openshift-servicemesh-plugin/issues/34

# Before
Screenshot from 2022-06-28 18-13-37

# After
Screenshot from 2022-06-28 19-33-44


aljesusg:
At least for init this and remove the error. We don't use i18n in kiali so It make no sense to have this configuration

jmazzitelli:
I see `docker-endpoint.sh` in that "Before" log - that is old - we don't have that anymore - the new image using the new base image no longer uses that. You should rebuild the server with the latest Dockerfile.

jmazzitelli (approved):
ran this and it works

PR #41, opened 2022-06-30
WIP ServiceMesh tab links

lucasponce:
WIP for https://github.com/kiali/openshift-servicemesh-plugin/issues/21


jmazzitelli (approved):
This works for me. I can see the links going to the correct place.

However, I am sure I am not testing every single link there is that needs to be tested. We are going to need a specific list of all the links in all the pages that need to be tested to confirm they work. QE is going to need something like this (or, better yet, have a Cypress test that clicks every link that has a "parent kiosk" and confirm it works).

lucasponce:
>  (or, better yet, have a Cypress test that clicks every link that has a "parent kiosk" and confirms it works).

+1

Yes, the functional side of the plugin is reaching a nice point, so in the next iterations, we can start adding them.

We can start adding them in a manual setup (to help a reviewer to test everything) and we can see how to automate it later (I don't think we can put a CRC in Github Actions, so that aspect will have a separated effort).

PR #48, opened 2022-07-27
Operator nightly and release pipeline

leandroberetta:
This PR adds the logic to the existing pipelines to release a nightly and a minor/patch versions of the operator in the same existing pipelines for the plugin.

As both components are in the same repo, the same pipelines will release a nightly and a release, tagging and commiting changes for next iteration.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -16,10 +16,15 @@ on:
         required: true
         default: main
         type: string
-      quay_repository:
-        description: Quay repository
+      plugin_quay_repository:
+        description: Plugin Quay repository
         type: string
         default: quay.io/kiali/servicemesh-plugin
+        required: true
+      operator_quay_repository:
+        description: Operator Quay repository
+        type: string
+        default: quay.io/kiali/ossmplugin-operator
         required: true
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -204,12 +224,20 @@ jobs:
...
+
     - name: Configure git
       run: |
         git config user.email 'kiali-dev@googlegroups.com'
 
         git config user.name 'kiali-bot'
 
+    - name: Create new version bundle (OLM)
+      run: |
+        ./operator/manifests/create-new-version.sh --new-version ${RELEASE_VERSION:1} --channels candidate
```
jmazzitelli, line 237: In order to use this `create-new-version.sh` you need to first download the operator SDK. See: https://github.com/kiali/openshift-servicemesh-plugin/blob/main/operator/manifests/create-new-version.sh#L42

So before this line you need to do `make get-operator-sdk`.
```suggestion
        make get-operator-sdk

        ./operator/manifests/create-new-version.sh --new-version ${RELEASE_VERSION:1} --channels candidate
```

jmazzitelli (changes requested):

.github/workflows/release.yaml
leandroberetta: It's true, thanks.

jmazzitelli (approved):

PR #54, opened 2022-08-08
Bump dynamic-plugin-sdk 0.0.13 + deps

lucasponce:
Update of "dynamic-plugin-sdk" to the latest version that incorporates new features around the YAMLEditor in plugin code.

This bump should be transparent to the current code, so a positive smoke test should be ok for validating it.


jmazzitelli (approved):
I ran a quick test to confirm it builds and deploys successfully. Make sure you check out the PR in your local repo, install Istio and Kiali, and then run the instructions found here to get this new plugin installed: https://github.com/kiali/openshift-servicemesh-plugin/blob/main/README.md#quick-summary

All works.

aljesusg (approved):
LGFM

PR #58, opened 2022-08-10
Add OpenShift doc link to the README

lucasponce:
Minor update of the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,8 @@ These are the things you need before you can start working with the OpenShift Se
 3. `oc` client available in the path
 4. `podman` client available in the path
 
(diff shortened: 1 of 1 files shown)


josunect (approved):
LGTM

I guess the plugin is based in the 4.10 version and the link does not need to be updated to the latest, right?

lucasponce:
> I guess the plugin is based in the 4.10 version and the link does not need to be updated to the latest, right?

Yes, it's mostly developed against 4.10. It should be compatible with 4.11 without modifications but has not yet been tested there.
In any case, according the documentation it would remain Tech Preview as well (meaning that changes are expected and it's still of an experimental nature).

Issue #67, opened 2022-08-26
new cr settings are not always properly processed


jmazzitelli:
I just noticed a problem in the new code in the operator that handles the new `kiali` settings.

```
    message: |
      The task includes an option with an undefined variable. The error was: 'kiali_service_namespace' is undefined

      The error appears to be in '/opt/ansible/roles/default/ossmplugin-deploy/tasks/main.yml': line 153, column 3, but may
      be elsewhere in the file depending on the exact syntax problem.

      The offending line appears to be:
      - set_fact:
```

This happens if you only set SOME but not ALL of the CR's kiali settings (e.g. my CR looked like this, and it caused the operator to abort with that error above):
```
  kiali:
    serviceName: ""
    serviceNamespace: foo
    servicePort: 0
    url: ""
```

I'll add some more code here to make sure this doesn't happen.

PR #68, opened 2022-08-26
Fix bug that aborts operator if you don't specify all of the kiali section

jmazzitelli:
This also optimizes route query - no need to query for the route 4 times, just get it once.

fixes: https://github.com/kiali/openshift-servicemesh-plugin/issues/67


jmazzitelli:
the molecule tests pass. and my manual testing also pass. this is good.

PR #69, opened 2022-08-26
rename version to 0.1.0

jmazzitelli:


jmazzitelli:
image

jmazzitelli:
there is no way to test pulling from quay - the make targets all assume the images the user wants come from the internal registry. I would like to be able to tell make to pull from quay. I'll add that to this PR.

jshaughn (approved):
LOK

Issue #52, opened 2022-08-08
Add Kiali validations on Istio Detail pages 


lucasponce:
In the current state of the plugin, the Istio "details" page is not implemented by the plugin, it's redirected to the standard page that OpenShift Console uses to visualize and edit any resource.

~~The ideal situation would be that the "dynamic-plugin-sdk" allow adding new tabs where additional "Service Mesh" information (like validations) can be added, but it seems this "extension point" is not yet available.~~

Then, the plugin should provide an implementation of a similar page that would allow augmenting the information with the Kiali validations, plus the standard edition features on the resource.

lucasponce:
Note, that now the plugin can add "Service Mesh" tab to Istio resources in a native way:
![image](https://user-images.githubusercontent.com/1662329/183883671-de25f9b3-1750-42b4-b189-f48b5666224c.png)

This issue can be used to design how to implement the "Service Mesh" tab at this level. 

One initial suggestion could be to delegate the CRUD capabilities to the original Openshift page, and the "Service Mesh" tab can query Kiali to collect the validations and show an editor with the validations in-line the sidepanel that Kiali standalone offers.

I think that adding new side panels to the original YAML editor is not yet available in the "dynamic-plugin-sdk".

hhovsepy:
Added Istio config details page in kiosk mode as read only YAML editor with validations in an iFrame contant:
![Screenshot from 2022-11-16 12-05-13](https://user-images.githubusercontent.com/604313/203630671-b1bde573-b5b0-4e75-ba6f-c698ea712690.png)

PR #72, opened 2022-08-31
Config details: Initial commit

hhovsepy:
RFE https://github.com/kiali/openshift-servicemesh-plugin/issues/52

### DIFF
--- a/plugin/package.json
+++ b/plugin/package.json
@@ -18,7 +18,8 @@
     "lint": "eslint ./src --fix"
   },
   "dependencies": {
-    "axios": "0.27.2"
+    "axios": "0.27.2",
+    "typestyle": "2.0.4"
   },
   "devDependencies": {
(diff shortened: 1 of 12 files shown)


aljesusg:
I would create a new repo with the components that we are using in kiali and openshiftserviceMesh... this will be hard of maintain and align with kiali/kiali.

israel-hdez:
I agree with @aljesusg . Copying and pasting code does not seem the way to go.
Creating yet another repo with re-usable components is OK. Although, I have some faith that the front-end in kiali/kiali can receive some moderate refactorings to make it re-usable (avoiding a new repo).

jshaughn:
In my experience it's pretty hard to build a good set of reusable components.  Although this may be a good idea, I think it's premature and I'd like to see us gain adoption for the the plugin, get feedback, and have a solid understanding of what we need to do to get it accpeted into console for real.  At that point we will be in a better position to decide on the steps necessary to maintain a solid code-base.

PR #78, opened 2022-09-08
plugin multi-arch build target and prepare for first release

jmazzitelli:
After you build the plugin itself, you should be able to run this new target to build and push the multi-arch containers.


jmazzitelli:
Note that I do not have `docker buildx` so I have been unable to test this.

jmazzitelli:
to test, you should just have to do this:

```
make build-plugin-image build-push-plugin-multi-arch
```

jmazzitelli:
I'm trying to test on my fork but it is failing due to some strange network error. I don't know if it is me or if it is github.

make/Makefile.plugin.mk
```diff
@@ -20,7 +20,7 @@ clean-plugin:
 
 ## build-plugin: Builds the plugin.
 build-plugin:
-	cd ${PLUGIN_DIR} && yarn install && yarn build
+	cd ${PLUGIN_DIR} && yarn install --network-timeout 3600000 && yarn build
```
nrfox, line 23: Should we add this to the dockerfile as well? https://github.com/kiali/openshift-servicemesh-plugin/blob/main/plugin/Dockerfile#L8

nrfox (approved):
Worked when testing locally. I can see all the arch images in my repo.

jmazzitelli:
After adding network timeout option to the yarn install cmd, it works in github actions.

jmazzitelli:
one final test is being performed - I'm testing the release workflow to ensure it releases multi-arch images (i.e. looking to see it perform a minor release with the correct version in the container image tag)

leandroberetta (approved):

PR #79, opened 2022-09-08
make it clear there are two options of installing and uninstalling

jmazzitelli:
The reader only needs to pick one of the methods, but we want to document both methods. I want to make it clear to the reader that two methods are being described, but you only need to pick one of the methods.


jshaughn (approved):
Dandy

PR #81, opened 2022-09-09
do not specify old version of NONE

jmazzitelli:
Only merge this AFTER we release the first (initial) release of the plugin.

For the first release, we need to specify NONE for the old version, but thereafter, there will always be an old version. This PR removes that option to say there is no old version - that allows the script to auto-discover what the last old version is (which is what we want for all subsequent releases).

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -236,8 +236,7 @@ jobs:
       run: |
         make get-operator-sdk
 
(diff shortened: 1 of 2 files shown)


jmazzitelli:
once https://github.com/redhat-openshift-ecosystem/community-operators-prod/pull/1613 is merged and the operator metadata is published, we can merge this PR.

Issue #87, opened 2022-09-12
plugin location hardcoded to ossmconsole


jmazzitelli:
The plugin location is hardcoded to be expected to be `ossmconsole`. That is wrong. We need to fix this:

https://github.com/kiali/openshift-servicemesh-plugin/blob/v0.1.0/operator/roles/default/ossmconsole-deploy/templates/openshift/consoleplugin.yaml#L10

jmazzitelli:
I do not see this in my own dev build. So something is breaking the plugin either in the CI workflow build OR in the way it is deployed via OLM (perhaps some security permissions thing? I did notice the Install Guide ends up installing the plugin pod in the openshift-operators ns and the dev makefile put it in ossmconsole - perhaps something is up with the way internal ns are treated?)

jmazzitelli:
```
$ oc logs -n istio-system -l app=ossmconsole
OpenShift Service Mesh Console: Version=[], Commit=[]
```

Not due to the namespace - even in istio-system it doesn't work.

Looks like the build is bad.

jmazzitelli:
I even switched to using the images I pushed to the internal registry (that I know worked) and I still can't get the plugin to work:

```
$ oc get pods -l app=ossmconsole -n istio-system -o jsonpath={.items[0].spec.containers[0].image} ; echo
image-registry.openshift-image-registry.svc:5000/kiali/ossmconsole:dev

$ oc get pods -l app=ossmconsole-operator -n openshift-operators -o jsonpath={.items[0].spec.containers[0].image} ; echo
image-registry.openshift-image-registry.svc:5000/kiali/ossmconsole-operator:dev
```

This images work when I deploy using the make targets (which create a OLM catalog, so it isn't like we are completely different - its still OLM).

There might be some differences in the metadata that causes the problem.

jmazzitelli:
This is how I tested the dev images but installing via community OLM (I need to point to the internal image registry):

```
OPERATOR_NAMESPACE="openshift-operators"
ENV_NAME="ALLOW_AD_HOC_OSSMCONSOLE_IMAGE"
ENV_VALUE="true"
[...]

PR #89, opened 2022-09-12
Update console service ref to same namespace as console

nrfox:
The ossmconsole Service is deployed in the same namespace as the ossmconsole resource and should use that value instead of hardcoding `ossmconsole`.

### DIFF
--- a/operator/roles/default/ossmconsole-deploy/templates/openshift/consoleplugin.yaml
+++ b/operator/roles/default/ossmconsole-deploy/templates/openshift/consoleplugin.yaml
@@ -7,7 +7,7 @@ spec:
   displayName: "OpenShift Service Mesh Console"
   service:
     name: ossmconsole
-    namespace: ossmconsole
+    namespace: {{ ossmconsole_vars.deployment.namespace }}
     port: 9443
     basePath: "/"
   proxy:


nrfox:
Fixes #87

jmazzitelli:
this looks correct. I will test now

jmazzitelli (approved):
This works. The only caveat is you need to set this in the CR:

```
  deployment:
    imageVersion: v0.1.0
```

This ensured it pulled the plugin image from quay (I didn't test with a local built version of the plugin).

Note I still see 
```
OpenShift Service Mesh Console: Version=[], Commit=[]
```

so that still needs to be fixed.

nrfox:
> Note I still see
>
>> OpenShift Service Mesh Console: Version=[], Commit=[]
>
> so that still needs to be fixed.

@jmazzitelli looks like that comes from the dockerfile as a build arg: https://github.com/kiali/openshift-servicemesh-plugin/blob/main/plugin/Dockerfile#L23 and needs to be passed in by CI when the image is built.

jmazzitelli:
> @jmazzitelli looks like that comes from the dockerfile as a build arg: https://github.com/kiali/openshift-servicemesh-plugin/blob/main/plugin/Dockerfile#L23 and needs to be passed in by CI when the image is built.

https://github.com/kiali/openshift-servicemesh-plugin/issues/90

Issue #90, opened 2022-09-12
missing version and commit info in log output


jmazzitelli:
I see this in the logs when installing via community OLM:

```
OpenShift Service Mesh Console: Version=[], Commit=[]
```

Looks like our CI builds are some how not providing that information. We need to fix that.

PR #92, opened 2022-09-12
pass proper build args when doing multi-arch build for plugin

jmazzitelli:
fixes: https://github.com/kiali/openshift-servicemesh-plugin/issues/90


nrfox (approved):
I can see the commit and version get propagated in the dockerfile build now:
```
RUN printf "\necho 'OpenShift Service Mesh Console: Version=[v0.2.0-SNAPSHOT], Commit=[35fbe020bcc4de326
```

PR #95, opened 2022-09-14
bump operator base image

jmazzitelli:
we are doing the same for kiali. See: https://github.com/kiali/kiali/issues/5459


jmazzitelli:
Test passes:

`make -e MOLECULE_SCENARIO="config-values-test" cluster-push operator-create molecule-test`

PR #102, opened 2022-11-07
Istio Config Details Service Mesh tab iFrame loads only config page with validations

hhovsepy:
RFE https://github.com/kiali/openshift-servicemesh-plugin/issues/52

Istio config objects which have validations in Kiali side now have "Service Mesh" tab in Console.

Screenshot from 2022-11-08 13-44-37

### DIFF
--- a/plugin/console-extensions.json
+++ b/plugin/console-extensions.json
@@ -165,5 +165,140 @@
       },
       "component": { "$codeRef": "MeshTab" }
     }
+  },
+  {
+    "type": "console.tab/horizontalNav",
+    "properties": {
+      "model": {
+        "group": "networking.istio.io",
+        "kind": "DestinationRule",
+        "version": "v1beta1"
+      },
+      "page": {
+        "name": "Service Mesh",
+        "href": "ossmconsole"
+      },
+      "component": { "$codeRef": "MeshTab" }
+    }
+  },
+  {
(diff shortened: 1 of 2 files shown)


josunect:
I think the view is much more integrated now. Some comments that I can think about: 

1. The Envoy filter or the HttpRoute doesn't have a Mesh Tab. I guess this is because they don't have validations? 

image
image

2. As the first impression the view is a bit confusing because you can see the same information in both tabs. I was wondering if indicating something more explicit like "Validations" can be more useful: 

image

image

hhovsepy:
Now in "Service Mesh" tab, the "YAML" subtab is renamed to "Validations", to avoid the confusion with the original "YAML" tab of console above.
Screenshot from 2022-11-10 09-51-28

Hi @andrew-ronaldson , wdyt from UX perspective about the "Service Mesh" tab?
It shows the Kiali Istio Config page with validations but in read-only mode.

josunect:
Reviewed this in the UX meeting sync. The proposal is, if we are going to have only one tab, to remove the "Validations" subtab, and update the name in the "Service Mesh"

hhovsepy:
Approached comments, thanks for review:
Screenshot from 2022-11-16 12-05-13

josunect (approved):
Tested and LGTM

Issue #73, opened 2022-08-31
[dev] instructions for running locally does not work


jmazzitelli:
I'm trying to follow these instructions: https://github.com/kiali/openshift-servicemesh-plugin#how-to-run-the-plugin-for-local-development

Both commands work and I see things running on ports 9000 (the plugin) and 9001 (the console) and I see the console in my browser.

However, the plugin is not visible in the console.

I see these errors in the console logs.

First I see a bunch of these:

```
2022/08/31 13:16:44 CheckOrigin: Proxy has no configured Origin. Allowing origin [http://localhost:9000] to wss://api....my-url-here....com:6443/apis/config.openshift.io/v1/clusterversions?watch=true&fieldSelector=metadata.name%3Dversion
```

Then this:

```
E0831 13:14:30.320303       1 handlers.go:165] GET request for "ossmconsole" plugin failed: Get "http://host.containers.internal:9001/plugin-manifest.json": dial tcp 172.17.0.1:9001: connect: connection refused
```

So something else is missing in the instructions because there seems to be a step that I need to do in order to get this to work.

jmazzitelli:
I thought maybe this would fix it (and I have a feeling we will still need this), but it still doesn't work when I do this:

```diff
diff --git a/plugin/plugin-config.json b/plugin/plugin-config.json
index 42a5276..9998e25 100644
--- a/plugin/plugin-config.json
+++ b/plugin/plugin-config.json
@@ -1,3 +1,6 @@
 {
-  "kialiUrl": "https://kiali-istio-system.apps-crc.testing"
-}
\ No newline at end of file
+  "kialiUrl": "https://kiali-istio-system.apps-crc.testing",
+  "serviceName": "kiali",
+  "serviceNamespace": "istio-system",
+  "servicePort": "20001"
+}
```

josunect:
I've have it running following the next steps:

- cd plugin
- yarn install
- cp plugin-config.json dist

Edited dist/plugin-config.json:

```
{
  "kialiUrl": "http://localhost:3000"
}
```

- yarn run start
- cd plugin
- yarn run start-console

![image](https://user-images.githubusercontent.com/49480155/187693074-2aa55835-e58d-48b2-aef7-ea16c49a1d31.png)

I can see  the following messages in the console: 

[...]

PR #106, opened 2023-02-16
Update start-console

aljesusg:
Fix https://github.com/kiali/openshift-servicemesh-plugin/issues/73

This update from https://github.com/openshift/console-plugin-template/blob/main/start-console.sh fix my local env.

### DIFF
--- a/plugin/start-console.sh
+++ b/plugin/start-console.sh
@@ -24,12 +24,18 @@ echo "API Server: $BRIDGE_K8S_MODE_OFF_CLUSTER_ENDPOINT"
 echo "Console Image: $CONSOLE_IMAGE"
 echo "Console URL: http://localhost:${CONSOLE_PORT}"
 
+
+# Prefer podman if installed. Otherwise, fall back to docker.
 if [ -x "$(command -v podman)" ]; then
-    echo "Using Podman"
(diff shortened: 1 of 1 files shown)


jmazzitelli:
This "works" - I do not get that error anymore. I do see the plugin in the local console:
image

BUT! I cannot see Kiali UI. When I try to log in, I get this:
image

I can confirm I am able to log in the Kiali route directly and see the Kiali UI. So the standalone Kiali is installed and working. I just can't see it in the plugin because it always asks me to log in, but the log in fails. This log-in issue is a known  problem, but in the past, we just had to log into the standalone Kiali and our login token will remain and the plugin never asked to log in again. It looks like this behavior is different when running in this local mode.

We can merge this PR though. It think this login issue is different than the network issue this PR fixed.

jmazzitelli (approved):
see my notes about the additional problem I see. But we can merge this PR since that is a different problem.

aljesusg:
Well. I can see all but I am running ./kiali/hack/run-kiali.sh that exposed kiali locally without authentication.
My next steps is move this to native and fix the authentication. But we can use the local env

josunect (approved):
Works as expected: 
image

image

(Tested in crc 2.13)

PR #125, opened 2023-03-27
the --to option no longer writes to stdout so we need to refactor this.

jmazzitelli:
part of https://github.com/kiali/kiali/issues/5945


ferhoyos (approved):
Tested and code LGTM.

ferhoyos (changes requested):
I have just noticed that same change needs to be applied to line 51 too.

jmazzitelli:
> I have just noticed that same change needs to be applied to line 51 too.

@ferhoyos this is done

ferhoyos (approved):
LGTM

PR #128, opened 2023-03-28
Include install-crd in the README

### LINKED ISSUE #127: No install-crd reference on README
I have followed https://github.com/kiali/openshift-servicemesh-plugin#building-and-deploying-the-operator to build and deploy the operator. At the end of `make cluster-push operator-create install-cr` the following message appears:

```
The CRD does not exist. You should install the operator.
```

Looking at the code, first stage of `install-cr` command is to wait for crd, but no crd is installed previously. I have seen another make command called `install-crd`, so probably the make command to execute is:

```
make cluster-push operator-create install-crd install-cr
```

ferhoyos:
Closes #127

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Here's a tl;dr summary to get the operator and plugin installed in your cluster.
 
 1. First run `make cluster-status` to expose the internal image registry and get the podman command needed to log into the internal image registry.
 2. Run the podman login command that will log into the internal image registry.
-3. Build, push, and deploy the operator and plugin by running `make cluster-push operator-create install-cr`
+3. Build, push, and deploy the operator and plugin by running `make cluster-push operator-create install-crd install-cr`
 
 When you are finished and you want to uninstall the operator and plugin, run `make operator-delete`.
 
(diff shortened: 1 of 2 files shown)


jmazzitelli:
i am pretty sure there is a reason I did not do this when I originally wrote that. Let me look and see why. Do not merge yet.

make/Makefile.operator.mk
```diff
@@ -131,7 +131,7 @@ purge-all-crs: .ensure-oc-login
 	  done
 
 ## run-operator: Runs the OSSM Console Operator via the ansible-operator.
-run-operator: install-crd install-cr get-ansible-operator
```
jmazzitelli, line 134: See, here is where you need to force the CRD because this task DOES install/run the operator so the CRD is needed. That's why this is here.

jmazzitelli (changes requested):
My opinion - just fix the docs to tell the user to run the `install-crd` target. See my comments.

README.md
```diff
@@ -71,7 +71,7 @@ Here's a tl;dr summary to get the operator and plugin installed in your cluster.
 
 1. First run `make cluster-status` to expose the internal image registry and get the podman command needed to log into the internal image registry.
 2. Run the podman login command that will log into the internal image registry.
-3. Build, push, and deploy the operator and plugin by running `make cluster-push operator-create install-cr`
+3. Build, push, and deploy the operator and plugin by running `make cluster-push operator-create install-crd install-cr`
```
jmazzitelli, line 74: Hmm. that's interesting. I thought the operator-create target does the install-crd for you. Now I see why you were confused... I guess I forgot to put that in the docs because I must have thought it did it already.

jmazzitelli (approved):

PR #130, opened 2023-03-29
fix the typo in the new ansible setting name

jmazzitelli:
The new setting name is `callbacks_enabled` (in the plural).

Also keep the deprecated name around since currently we are using an ansible version that doesn't know about the new one.

part of https://github.com/kiali/kiali/issues/4338


jshaughn (approved):
LOK

Issue #144, opened 2023-04-26
[operator] remove ansible loops for better performance


jmazzitelli:
similar to: https://github.com/kiali/kiali/issues/6032

we need to remove ansible loops where possible since they degrade performance alot.

PR #145, opened 2023-04-26
[operator] remove ansible loops for better performance

jmazzitelli:
fixes: https://github.com/kiali/openshift-servicemesh-plugin/issues/144

To test this, just push the dev images to the cluster and install the plugin via the operator. It should still work.

```
make cluster-push operator-create install-crd install-cr
```


ferhoyos (approved):
Tested, operator installs plugin correctly.

PR #155, opened 2023-06-05
Use same prettier config as Kiali and core-ui projects

jshaughn:
Hi @ferhoyos , I noticed a lot of prettier warning when I created an initial vscode project for ossmc.  Not sure if it was intentional to have a different prettier config for OSSMC, but if not let's use the same as Kiali and core-ui.  Otherwise we should do a big reformat of the OSSMC repo.

### DIFF
--- a/plugin/.prettierrc.json
+++ b/plugin/.prettierrc.json
@@ -1,0 +1,6 @@
+{
+  "arrowParens": "avoid",
+  "printWidth": 120,
+  "singleQuote": true,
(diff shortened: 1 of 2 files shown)


ferhoyos:
To be consistent, I think we should use same prettier config in all kiali repositories. Even we should use same prettier file format (I don't know why kiali upstream use json format but OSSMC prettier file is defined as yaml).

ferhoyos (approved):
LGTM

PR #156, opened 2023-06-06
Move ProjectMesh tab to native

### LINKED ISSUE #151: MeshTab Project View
Move ProjectView that we see in the MeshTab to native

ferhoyos:
Closes #151.


jshaughn:
Hi @ferhoyos , I'd like to have a meeting to go over this approach with myself and the greater team.  But let's also start capturing questions and feedback here in the PR...

I could not get the PR in it's current state to run, so I'm not sure if that's expected.  Maybe it needs a rebase.  I get the same results as the failures above, in the CI.

From what I can tell, the approach for this PR is to literally copy nearly all of the kiali/frontend into this repo and then use it to back a fairly thin facade for the actual plugins.  In that case I wonder if we could almost generate a "versioned Kiali frontend", almost as one big component.  Anyway, tomorrow let's talk more...

ferhoyos:
Hi @jshaughn, the CI tests are failing because @kiali/types version used here is not available in npm yet. In local development you need to use following PR of core-ui (https://github.com/kiali/core-ui/pull/42) and link the library to the plugin (as explained in the core-ui README).

You are right, as first iteration the idea is:
-  Business logic in @kiali/types library - same reusable code in all kiali applications
- Visual components (pages and components) copied to plugin and from here analyze what is needed to make plugin become GA (e.g., links and styles). I have followed the fairly thin facade approach to make maintenance and new feature transition from upstream to plugin easier (most of the code will be the same).

Once we analyzed what is needed to make the plugin GA, if these changes are compatible with upstream we could think of generating a "versioned Kiali frontend" in future iterations to have only one Kiali source code. Currently there are parts of the application implemented just for upstream (e.g. hardcoded links in some components), first we should make them generic enough to make it work for both scenarios. Another alternative is to analyze which components are exactly the same in both scenarios and move it to core-ui (first small components, like ValidationSummary). The idea here is to have as less duplicated code as possible.

ferhoyos:
Closing since issue #151 is closed.

Issue #166, opened 2023-06-26
Create README file with Kiali copy information


ferhoyos:
Issue created from https://github.com/kiali/openshift-servicemesh-plugin/pull/162#issuecomment-1607303266. It consists on modyfing the copy script to generate programmatically a README file in `plugin/src/kiali` folder that contains Kiali copy information.

An example of this README could be:
```
**WARNING**: Code in this folder comes from kiali/kiali repository and it should never be modified here.

Kiali frontend source code information:

- git ref: v1.70
- git commit: 1d0d4d1ab62d4c0b9d9ce0f4d310d7bb4044d4ff
- GitHub URL: https://github.com/kiali/kiali/tree/1d0d4d1ab62d4c0b9d9ce0f4d310d7bb4044d4ff/frontend/src
```

I think it is better a README than javascript file because this information can be read from github if you access to `plugin/src/kiali`folder

PR #167, opened 2023-06-26
[hack] have script update README

jmazzitelli:
fixes: https://github.com/kiali/openshift-servicemesh-plugin/issues/166


ferhoyos (approved):
I have tested the script and README file is generated correctly.

Issue #28, opened 2022-06-17
plugin needs to know the kiali.url for the specific tenant it wants to observe


jmazzitelli:
When only one Kiali is installed in the cluster (i.e. there is no multi-tenancy - all namespaces in the cluster are accessible to a single instance of Kiali), then the plugin can use the kiali.url setting that the user provided in the OSSMPlugin CR (when left empty, the operator will set this for the user - which will be the typical use case). This means the one Kiali route instance will be where that kiali.url setting points to.

However, this breaks down in multi-tenant scenarios. There isn't just one kiali.url the plugin needs to know about. Based on what namespace the plugin wants to observe, it has to figure out which Kiali route URL to talk to.

This is relatively easy to accomplish because the Kiali operator adds a label to all accessible namespaces - `kiali.io/member-of`. The value of that namespace label will be the name of the control plane of the namespace's tenant - and that control plane is where the Kiali route can be found.

So, for example, if I have an application deployed in the `alpha` namespace and that namespace is part of a tenant controlled by that tenant's control plane, this is how we can find out which Kiali is observing that tenant:

(note: I am using Bash commands here, but obviously, this is just pseudo-code - we'd have to do this in the plugin javascript or other code such as that, rather than rely on Bash)

First, get the label to find out which control plane namespace where the observing kiali lives in:

```bash
KIALI_NAMESPACE="$(oc get namespace alpha -o jsonpath="{.metadata.labels['kiali\.io/member-of']}")"
```

Now that we know the control plane namespace, we know where the observing Kiali lives (it is in that control plane namespace itself). Thus, we just look in that namespace for the Route (the control plane namespace is where the Kiali routes lives, too):

```bash
KIALI_ROUTE="https://$(oc get routes -n ${KIALI_NAMESPACE} -l app.kubernetes.io/name=kiali -o jsonpath="{.items[0].spec.host}")"
```

The $KIALI_ROUTE is now the kiali.url that the plugin should use - that is the Kiali that the plugin should be talking to when looking at workloads within that namespace.

The OSSM Plugin CR also has `serviceName`, `serviceNamespace`, and `servicePort` in its `kiali` section, but all of those can be obtained from the route that is auto-discovered:

```bash
SERVICE_NAME="$(oc get route -n ${KIALI_NAMESPACE} -l app.kubernetes.io/name=kiali -o jsonpath={.items[0].spec.to.name})"
SERVICE_NAMESPACE="${KIALI_NAMESPACE}"
[...]

PR #168, opened 2023-06-27
Replace iframes with wrappers

### LINKED ISSUE #160: Replace iframes with wrappers
React components that implement current iframes must be replaced by wrappers that use Kiali application copied from [#kiali/kiali](https://github.com/kiali/kiali). Besides, a controller has to be created to load initial configuration from Kiali backend server (similar functionality provided by AuthenticationController in Kiali application).

Wrappers to be implemented:
- Overview page
- Graph page
- IstioConfigDetails page
- ServiceDetails page
- WorkloadDetails page

ferhoyos:
Closes #160


plugin/manifest.yaml
```diff
@@ -104,23 +104,6 @@ data:
...
-  labels:
-    app: ossmconsole
-    app.kubernetes.io/component: ossmconsole
-    app.kubernetes.io/instance: ossmconsole
-    app.kubernetes.io/name: ossmconsole
-    app.kubernetes.io/part-of: ossmconsole
-data:
-  plugin-config.json: |
-    {
-      "kialiUrl": "https://kiali-istio-system.apps-crc.testing"
```
jmazzitelli, line 120: I see the `kialiUrl` is being removed in a bunch of places in this PR.

Is `kialiUrl` going to go away completely? If so, a bunch of changes need to be made to the operator ([here](https://github.com/kiali/openshift-servicemesh-plugin/blob/main/operator/roles/default/ossmconsole-deploy/templates/openshift/configmap-plugin.yaml#L10) and related places)

ferhoyos: Yes, kialiUrl is not used since there are no iframes to redirect to kialiUrl. I will try to change as much as I can but probably I wll need your help on operator @jmazzitelli

jmazzitelli:
If the kialiUrl is getting removed, then this issue can be closed since it involves the kialiUrl setting:
* https://github.com/kiali/openshift-servicemesh-plugin/issues/28

plugin/manifest.yaml
jmazzitelli: BTW: if we get rid of the kialiUrl, how does the plugin know which backend Kiali server to talk to?

plugin/plugin-config.json
```diff
@@ -1,3 +0,0 @@
-{
-  "kialiUrl": "https://kiali-istio-system.apps-crc.testing"
```
jmazzitelli, line 2: My suggestion here is to just remove this line, but do NOT remove the entire file.

I would think chances are good that at some point in the future, we are going to need some ability to configure the  plugin via a config file. So let's leave the framework in place. Even if the config file is empty for now, let's leave it empty but keep it. So when the time comes when we need to add a new config option, we don't have to do the work all over again to add the config file.

Unless you know for sure we will never need a config file - if we won't ever need a config file, then we can remove it.

README.md
```diff
@@ -57,14 +56,8 @@ Alternatively, you can manually set up your dev environment outside of make by p
 cd plugin
 yarn install
 
-# If necessary, make sure you change the "KIALI_PROXY" value in .env.development so it points to your Kiali Server URL
+# If necessary, make sure you change the "API_PROXY" value in .env.development so it points to your Kiali Server URL
 # vi .env.development
-
-# Copy the plugin-config.json file into the "dist" folder to emulate the ConfigMap in a local environment
-cp plugin-config.json dist
```
jmazzitelli, line 64: I would leave this here - maybe add to the comment saying, `This config file is empty today so is not really needed. This is here for future enhancement.`

make/Makefile.plugin.mk
```diff
@@ -71,8 +71,7 @@ endif
 ## prepare-dev-env: Prepares the local dev environment so you can run the plugin and OpenShift console locally.
 prepare-dev-env: .determine-kiali-url
 	@cd ${PLUGIN_DIR} && yarn install
-	@cp ${PLUGIN_DIR}/plugin-config.json ${PLUGIN_DIR}/dist && sed -i -r 's|"kialiUrl": "(.*)"|"kialiUrl": "${KIALI_URL_TO_USE}"|' ${PLUGIN_DIR}/dist/plugin-config.json
```
jmazzitelli, line 74: Leave the copy of the config file, even if it is empty (for future enhancement)

plugin/manifest.yaml
```diff
@@ -104,23 +104,6 @@ data:
     }
 ---
 apiVersion: v1
-kind: ConfigMap
```
jmazzitelli, line 107: since this isn't needed today, it doesn't have to be created in the dev environment. But it doesn't hurt to create one that has empty data, just as a placeholder for future enhancement. Or comment this out. But I would leave it here in some state (commented out or not) just so in the future when we need to add configuration, we can just resurrect this without worrying about how it needs to be done. The work is already done here.

jmazzitelli (changes requested):
Unless we know for sure we will never need configurability in the plugin, I suggest we leave the ConfigMap/config file in place - even if it is empty right now. That way we don't have to repeat work that was already done in the future if we need to add a config option.

plugin/manifest.yaml
ferhoyos: kialiUrl is used just for iframes, backend kiali service is set in OSSMConsole CRD:

https://github.com/kiali/openshift-servicemesh-plugin/blob/057308b64bb57a9f1a11fb193fa87eb2802891ed/operator/crd-docs/crd/kiali.io_ossmconsoles.yaml#L85

jmazzitelli:
operator work to remove kiali.url is done here: https://github.com/ferhoyos/openshift-servicemesh-plugin/pull/4

plugin/manifest.yaml
jmazzitelli: Oh.... I see it... its the proxy definition in the ConsolePlugin object: https://github.com/ferhoyos/openshift-servicemesh-plugin/blob/main/operator/roles/default/ossmconsole-deploy/templates/openshift/consoleplugin.yaml#L13-L20

jmazzitelli:
now that the new frontend code is in master, I can't build this PR to test it... lots of errors about default exports. So I'll wait for updates to this PR before testing

example of error (there are many like this):
```
ERROR in ./src/openshift/utils/Reducer.ts:16:8
TS1192: Module '"/home/jmazzite/source/kiali/openshift-servicemesh-plugin/plugin/src/kiali/reducers/MessageCenter"' has no default export.
    14 | import tour from 'reducers/TourState';
    15 | import namespaces from 'reducers/NamespaceState';
  > 16 | import Messages from 'reducers/MessageCenter';
```

ferhoyos:
Compilation errors are fixed. @jmazzitelli you can test PR again.

jmazzitelli:
BTW... @ferhoyos while looking up how to install this via make, I noticed this in the README that is incorrect : https://github.com/kiali/openshift-servicemesh-plugin/pull/172

PR #169, opened 2023-06-29
remove kiali url

jmazzitelli:
@ferhoyos this is just a draft PR to show what needs to be done to remove the kiali.url configuration setting. We can merge this as-is or you can add this to your PR. If you prefer, I can create a PR to your PR with this information so it would be easier for you to integrate this in your work. Let me know what you want to do with this.

### DIFF
--- a/operator/crd-docs/cr/kiali.io_v1alpha1_ossmconsole.yaml
+++ b/operator/crd-docs/cr/kiali.io_v1alpha1_ossmconsole.yaml
@@ -20,4 +20,3 @@ spec:
     serviceName: ""
     serviceNamespace: ""
     servicePort: 0
-    url: ""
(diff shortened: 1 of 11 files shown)


ferhoyos:
I think it is better to create a PR to my PR, so I can check that plugin operator installation is ok with my PR code (except for the cluster param issue, but that's another story).

jmazzitelli:
I did run this and things works. It even was able to get the kiali version using the service directly (not the route).

```
$ kubectl get ossmconsoles.kiali.io ossmconsole -n ossmconsole -o jsonpath='{.status.environment.kialiVersion}{"\n"}'
v1.71.0-SNAPSHOT
```

jmazzitelli:
> I think it is better to create a PR to my PR, so I can check that plugin operator installation is ok with my PR code (except for the cluster param issue, but that's another story).

OK. Stay tuned for that - I'll create that shortly.

PR #172, opened 2023-07-03
do not tell user to install-crd - OLM is responsible for it and will install it

jmazzitelli:
The instructions are incorrect - the user should not "install-crd" here. When installing the operator via OLM, OLM is responsible for installing the CRD - when OLM deploys the operator, it will also deploy the CRDs associated with the operator.

Note that the `install-cr` make target will wait for the CRD to be deployed. So after `operator-create`, OLM will take some time but eventually deploy the operator and then its CRD. The `install-cr` make target will wait until that CRD is deployed before it creates the actual CR.


ferhoyos (approved):
I have test the process and there is no need to install-crd since it is deployed by OLM.

PR #174, opened 2023-07-05
Set PF Graph as default option in OSSMC

### LINKED ISSUE #173: Graph Selection in OSSMC plugin
Now that PFT graph has been added to Kiali code, it is time to be able to select between Cytoscape and PFT graph in the OSSMC plugin. PFT graph should be the default option.

ferhoyos:
Closes #173.

Testing PR:
Check that PF graph is always shown in OSSMC (although Kiali CR has Cytoscape graph option set):
- Graph section:
image

- Service detail
image

- Workload detail
image

- Project detail
image


jshaughn (commented):
This works but actually I'd like for OSSMC users to be able to switch to Cytoscape if desired.  Do you guys see any way we could do that?   If the PFT graph is not sufficient while we are improving it, I think it would be nice to allow a backdoor to Cy.

jmazzitelli:
> Do you guys see any way we could do that? If the PFT graph is not sufficient while we are improving it, I think it would be nice to allow a backdoor to Cy.

Yes - through the config file that I had @ferhoyos retain last week because [I said](https://github.com/kiali/openshift-servicemesh-plugin/pull/168#pullrequestreview-1505264952), `I suggest we leave the ConfigMap/config file in place - ... if we need to add a config option.` :)

So we can add the config option to the OSSMC CR which translates to putting the option in the config json that the plugin can use. For a dev, you can directly edit the ConfigMap that contains the config json.

e.g. we can add an option [here](https://github.com/kiali/openshift-servicemesh-plugin/blob/main/operator/crd-docs/cr/kiali.io_v1alpha1_ossmconsole.yaml#L19) in the OSSMC CR which then would go into the ConfigMap [here](https://github.com/kiali/openshift-servicemesh-plugin/blob/main/operator/roles/default/ossmconsole-deploy/templates/openshift/configmap-plugin.yaml#L10). That is empty now, but we used to put the URL in here - so to see what it would look like, see that [here](https://github.com/kiali/openshift-servicemesh-plugin/blob/v0.3/operator/roles/default/ossmconsole-deploy/templates/openshift/configmap-plugin.yaml#L10).

jmazzitelli:
IMO, anytime we use the hack script to copy the front end sourcecode to this repo, we should do that copy in a separate PR. Several reasons for that: (a) it separates it in a separate commit so it is easily seen in the history (b) when combined with other work, it looks like more files were touched to do that other work. For example, this PR says 75 were changed but only a handful were touched to incorporate the PF stuff into the plugin.

If we want to do the work and the hack-script-copy in a single PR, make sure we do NOT squash merge the PR. Keep the commits separate so we can see the hack-script-copy in history and we can see the commit message.

ferhoyos:
Closing since this PR has been split in #175 and #176 to ease the review.

PR #189, opened 2023-08-08
Copy of Kiali frontend source code (v1.72.0)

ferhoyos:
Kiali frontend source originated from:
* git ref:    v1.72.0
* git commit: 9c1cf48f8a032274d95e04d6976b7d2c26150420
* GitHub URL: https://github.com/kiali/kiali/tree/9c1cf48f8a032274d95e04d6976b7d2c26150420/frontend/src


plugin/src/openshift/components/KialiController.tsx
```diff
@@ -90,7 +90,7 @@ class KialiControllerComponent extends React.Component<KialiControllerProps> {
 
   render() {
     return this.state.configLoaded ? (
-      <div className={`${globalStyle} ${ossmcStyle} ${varStyle.kiali}`}>{this.props.children}</div>
+      <div className={`${globalStyle} ${ossmcStyle} ${cssVariables.style}`}>{this.props.children}</div>
```
jmazzitelli, line 93: Is this commit "change cssVariables name" something you have to do every time? If so, should we add this to the copy script so it isn't manual?

If this is just a one-off change, then never mind :)

jmazzitelli (approved):
Very nice, I see we used the script to pin to the 1.72.0 source as opposed to master.... LGTM

PR #190, opened 2023-08-10
no need for build-plugin-image to call build-plugin.

jmazzitelli:
The Dockerfile runs the yarn install in the container already. It is pointless to run it locally (which is what build-plugin does). We need the yarn install run inside the plugin during the docker build, which is what this still does.


ferhoyos (approved):
LGTM

PR #200, opened 2023-08-31
Set latest Kiali code with PF4 for OSSMC (v1.73.0)

ferhoyos:
Set latest Kiali code with PF4 for OSSMC (v1.73.0) as base code for future OSSMC GA version.


jmazzitelli (approved):
this works:
image

PR #201, opened 2023-08-31
workaround opm render bug by providing custom docker_config location

jmazzitelli:
part of: https://github.com/kiali/kiali/issues/6545

This is the same solution that was added to the kiali/kiali repo: https://github.com/kiali/kiali/pull/6546


ferhoyos (approved):
LGTM. Tested for OSSMC and it works too.

Issue #206, opened 2023-09-11
release workflow needs to update version in package.json


jmazzitelli:
This needs to be updated when the release workflow runs:

https://github.com/kiali/openshift-servicemesh-plugin/blob/v0.4.0/plugin/package.json#L133

See `version:`
```json
  "consolePlugin": {
    "name": "ossmconsole",
    "version": "0.1.0",
    "displayName": "OpenShift Service Mesh Console",
```

PR #207, opened 2023-09-11
update consolePlugin.version in package.json on release

jmazzitelli:
fixes: https://github.com/kiali/openshift-servicemesh-plugin/issues/206

This PR will update the release workflow so it automatically updates the plugin/package.json file's consolePlugin.version field (it was not getting updated previously - it is still stuck on 0.1.0).

This PR also manually updates the consolePlugin.version field so it is up to date (and matches the version field). This is just so it is consistent and matches the actual current version (0.5.0).  With the change in the release workflow that this PR is introducing, we will not have to manually update this consolePlugin.version field anymore.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -215,8 +215,9 @@ jobs:
         sed -i -r "s/^VERSION \?= (.*)/VERSION \?= $RELEASE_VERSION/" Makefile
 
         # UI version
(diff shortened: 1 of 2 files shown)


jmazzitelli:
tested on my fork and it works: https://github.com/jmazzitelli/openshift-servicemesh-plugin/actions/runs/6150912256/job/16689886141

jshaughn (approved):
LG.  If our next release is a GA, should we up the version to 1.0.0 (not now, but before release).

PR #226, opened 2023-12-05
updating cypress packages

ScriptingShrimp:
test of CI

### DIFF
--- a/plugin/package.json
+++ b/plugin/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ossmconsole",
-  "version": "1.77.0",
+  "version": "1.78.0",
   "private": true,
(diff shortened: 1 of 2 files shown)


ferhoyos (changes requested):
@ScriptingShrimp please commit yarn.lock file everytime you update package.json

plugin/package.json
```diff
@@ -43,6 +44,7 @@
     "lodash-es": "^4.17.21",
     "micro-memoize": "4.0.9",
     "moment": "^2.29.4",
+    "postcss": "^8.4.12",
```
ferhoyos, line 47: Why async and postcss libraries need to be added for cypress upgrade?

PR #228, opened 2023-12-11
update of cypress and testing libraries

ScriptingShrimp:
follow up of https://github.com/kiali/openshift-servicemesh-plugin/pull/226
@ferhoyos can you please verify this change does not brake anything?


plugin/cypress/support/step_definitions/duckduckgo.ts
ScriptingShrimp, line 1: For some reason new Cypress searches for step definitions in this directory, not sure why, have not yet find a reference. Its different from path than in kiali/kiali but it works for now. If we want to use exactly same structure as in Kiali, we can file the issue to address this.

ferhoyos: It seems that Cypress has changed a lot lately. Here new default paths of cypress tests and different possibilities to structure your e2e tests are explained:

https://github.com/badeball/cypress-cucumber-preprocessor/blob/master/docs/step-definitions.md

IMO I do not like path `support/steps_definitions`. I would choose the path `e2e` (even better that the one defined in Kiali - `integration`). How to distribute feature and .ts files inside this folder? It will depend how tests are related each other.

plugin/package.json
```diff
@@ -70,7 +71,8 @@
   "devDependencies": {
     "@babel/core": "^7.16.0",
     "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
-    "@badeball/cypress-cucumber-preprocessor": "^12.2.0",
+    "@badeball/cypress-cucumber-preprocessor": "^19.2.0",
+    "@bahmutov/cypress-esbuild-preprocessor": "^2.2.0",
     "@cypress/browserify-preprocessor": "3.0.2",
```
ferhoyos, line 76: We are using esbuild-preprocessor so you can remove browserify one from package.json

ferhoyos (changes requested):
Good job @ScriptingShrimp! I think you are in the right path starting with sample test to check that all infrastructure works fine (some changes with respect to older version like esbuild preprocessor and step definitions path). Check my comments and let's decide what is the best path approach for the moment (we can always change afterwards if we discover that this initial approach does not work).

plugin/cypress/support/step_definitions/duckduckgo.ts
ScriptingShrimp: I don't mind to have all stuff in one place, i.e.: if we have some simple scenarios with basic steps/not reusable steps, the `.js` and `.feature`  can be side by side for clarity. 

Shared step definitions is something we can break down more logically, specifically in kiali/kiali. Maybe we want to have `.js` file which contains only logic to handle table navigation, or graph manipulation. This could be more readable for us: straightforward to navigate to `shared/` and there you can find sorted methods of your interest.

plugin/package.json
ScriptingShrimp: @ferhoyos I removed `@cypress/browserify-preprocessor` package

ScriptingShrimp:
pushed few changes to address comments above

plugin/cypress/e2e/duckduckgo.ts
```diff
@@ -0,0 +1,13 @@
+import { When, Then } from "@badeball/cypress-cucumber-preprocessor";
+
+When("I visit duckduckgo.com", () => {
+  cy.visit("https://www.duckduckgo.com");
+});
+
+Then("I should see a search bar", () => {
+  cy.get("input").should(
```
ferhoyos, line 8: The test does not work for me, I have fixed it adding id to input 
```
cy.get('input#searchbox_input')
```

ScriptingShrimp: For Jenkins integration it's better to have it  failing for now. It's just dummy test, will be removed in next PR coming soon.

ferhoyos (approved):
LGTM

PR #230, opened 2023-12-21
authorization for OSSMC tests + initial smokes 

ScriptingShrimp:
I was tinkering with authorization, but was unable to get it working using API - we can look into that later.

This PR introduces authorization for tests using new `cy.session` method, which can persist across multiple spec files. For initial testing this should be enough. 

Few tests to verify the routes are bundled as `@smoke` for OSSMC. There will be more smokes to verify that the actual UI of Kiali is loaded and displayed  correctly. Also there are few `TODO` comments we can adders later.


plugin/cypress/support/commands.ts
```diff
@@ -25,13 +25,29 @@
...
+    cy.visit('/').then(() => {
+        cy.log('OC_IDP: ', typeof (idp), JSON.stringify(idp))
+        if (idp != undefined) {
+            cy.get('.pf-c-button').contains(idp).click()
+        }
+        cy.get('#inputUsername').type(user)
+        cy.get('#inputPassword').type(password)
+        cy.get('button[type="submit"]').click();
+    })
+})
```
nrfox, line 53: What this does is login to openshift console, save the cookie with `cy.session`, then the console plugin uses that login cookie to authenticate with the kiali api? I think this is ok for the console plugin since I think the only auth strategy the console plugin will work with is openshift? Whereas standalone Kiali must with work a number of different auth strategies.

plugin/cypress/support/commands.ts
```diff
@@ -25,13 +25,29 @@
...
-//   }
-// }
\ No newline at end of file
+declare global {
+    namespace Cypress {
+        interface Chainable {
+            login(OC_CLUSTER_USER: string, OC_CLUSTER_PASS: string, OC_IDP: string): Chainable<void>
+            //   drag(subject: string, options?: Partial<TypeOptions>): Chainable<Element>
+            //   dismiss(subject: string, options?: Partial<TypeOptions>): Chainable<Element>
+            //   visit(originalFn: CommandOriginalFn, url: string, options: Partial<VisitOptions>): Chainable<Element>
```
nrfox, line 34: This just looks like boilerplate that can be removed.

nrfox (commented):
Looks like a good start! I haven't tested it but it looks fine.

plugin/cypress/support/commands.ts
ScriptingShrimp: That is a great question. I'm not aware if there are any restrictions regarding authorization for OCP to display OSSMC. For this case I followed the simplest scenario to authorize into the console and smoke test the integration which is essential for the release pipeline.

ScriptingShrimp: The `cy.session` usage is based on this [approach from the docs](https://docs.cypress.io/api/commands/session#Caching-session-data-across-specs) . All is done in the UI, there is room for improvements, but the most important thing is that the login routine is not triggered each scenario or spec file and it's cached for reuse. Additionally I want to implement session validation (see in the link above) so tests could be even faster (nice to have for now).

plugin/cypress/support/commands.ts
ScriptingShrimp: > This just looks like boilerplate that can be removed.

Will remove with the next commit.

plugin/cypress.config.ts
```diff
@@ -14,13 +14,21 @@ export default defineConfig({
   requestTimeout: 15000,
   responseTimeout: 15000,
   fixturesFolder: 'cypress/fixtures',
+  chromeWebSecurity: false,
+  screenshotsFolder: "results/screenshots",
+  videosFolder: "results/videos",
+  // videoUploadOnPasses: false, // this is not supported in cypress 13+, TODO add this back to config
+
   env: {
+    USERNAME: "jenkins", // default value for jenkins
```
ferhoyos, line 23: It is not consistent with username variable in the rest of the configuration (`OC_CLUSTER_USER`). In my opinion I think it is more simple to set just `OC_USERNAME` and `OC_PASSWORD` variable names.

plugin/cypress.config.ts
```diff
@@ -14,13 +14,21 @@ export default defineConfig({
...
+    OC_IDP: "my_htpasswd_provider", // default value for jenkins, can vary based on cluster setup
     'cypress-react-selector': {
       root: '#root'
     },
     omitFiltered: true,
     filterSpecs: true
   },
+
   e2e: {
     baseUrl: 'http://localhost:3000',
```
ferhoyos, line 33: Openshift CRC local base url is `http://localhost:9000`

plugin/cypress/support/commands.ts
```diff
@@ -8,30 +8,27 @@
...
+
+Cypress.Commands.add('login', (OC_CLUSTER_USER, OC_CLUSTER_PASS, OC_IDP) => {
+    const user = OC_CLUSTER_USER || Cypress.env('OC_CLUSTER_USER')
+    const password = OC_CLUSTER_PASS || Cypress.env('OC_CLUSTER_PASS')
+    const idp = OC_IDP || Cypress.env('OC_IDP')
+
+    cy.visit('/').then(() => {
+        cy.log('OC_IDP: ', typeof (idp), JSON.stringify(idp))
+        if (idp != undefined) {
+            cy.get('.pf-c-button').contains(idp).click()
```
ferhoyos, line 28: I guess that you have used standard Openshift to test PR. In my case I have used local CRC and it works in a different way:
- If you access CRC there is no idp button, direct username/password input types. We should add special idp to avoid line 28 (`undefined` is not possible since there is default value)
- If you deploy local CRC with `yarn start` command there is no login page (like `anonymous` option in Kiali). For this case I would add another special idp value (e.g., `anonymous`) to skip login process.

My 2cents for the method:
```
Cypress.Commands.add('login', (OC_USERNAME, OC_PASSWORD, OC_IDP) => {
  const user = OC_USERNAME || Cypress.env('OC_USERNAME');
  const password = OC_PASSWORD || Cypress.env('OC_PASSWORD');
  const idp = OC_IDP || Cypress.env('OC_IDP');

  if (idp !== 'anonymous') {
    cy.visit('/').then(() => {
      cy.log('OC_IDP: ', typeof idp, JSON.stringify(idp));
      if (idp !== 'crc') {
        cy.get('.pf-c-button').contains(idp).click();
      }
      cy.get('#inputUsername').clear().type(user);
      cy.get('#inputPassword').clear().type(password);
      cy.get('button[type="submit"]').click();
    });
  }
});
```

Probably `crc` is not the best idp. We should investigate how many idp options we could have in Openshift and test them.

ferhoyos (changes requested):
This is a great start @ScriptingShrimp! Now we can run test suites in Openshift. I have run tests successfully with local CRC after applying some changes, check my comments. 

One thing to define is the default values. In my opinion default values should be the ones used in local development (after executing `yarn start`), so developers do not need to export env variables before executing `yarn cypress:open` or `yarn cypress:run` commands. In automated tests we can define variables to export in the script.

plugin/cypress.config.ts
```diff
@@ -14,13 +14,21 @@ export default defineConfig({
   requestTimeout: 15000,
   responseTimeout: 15000,
   fixturesFolder: 'cypress/fixtures',
+  chromeWebSecurity: false,
```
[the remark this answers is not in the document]
ScriptingShrimp: I don't have, but I was considering this to be safe default for some reasons. I will remove it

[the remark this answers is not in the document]
ScriptingShrimp: enabled back on since we don't have any cross origin requests

plugin/cypress/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Prerequisites
+
+Installed all dev dependencies from frontend folder. Ensure the `baseUrl` field in the `cypress.config.ts` file is pointing to the server you are trying to test, alternatively you can set `CYPRESS_BASE_URL` environment variable or pass via cmd line `yarn cypress --config baseUrl=http://kiali-server:20001` to overwrite default `baseUrl`. 
+
+
+Before you start using Cypress suite, you might need export some environment variables:
+
+```bash
+export CYPRESS_BASE_URL=<value>               # defaults to http://localhost:3000
```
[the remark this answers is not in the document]
ScriptingShrimp: fixed

plugin/cypress/README.md
```diff
@@ -0,0 +1,30 @@
+# Cypress Visual Testing with BDD framework
+
+These are visual tests for kiali that are meant to be run against a live environment.
+
+## Prerequisites
+
+Installed all dev dependencies from frontend folder. Ensure the `baseUrl` field in the `cypress.config.ts` file is pointing to the server you are trying to test, alternatively you can set `CYPRESS_BASE_URL` environment variable or pass via cmd line `yarn cypress --config baseUrl=http://kiali-server:20001` to overwrite default `baseUrl`.
```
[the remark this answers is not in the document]
ScriptingShrimp: fixed

plugin/cypress/README.md
```diff
@@ -0,0 +1,30 @@
...
+## Prerequisites
+
+Installed all dev dependencies from frontend folder. Ensure the `baseUrl` field in the `cypress.config.ts` file is pointing to the server you are trying to test, alternatively you can set `CYPRESS_BASE_URL` environment variable or pass via cmd line `yarn cypress --config baseUrl=http://kiali-server:20001` to overwrite default `baseUrl`. 
+
+
+Before you start using Cypress suite, you might need export some environment variables:
+
+```bash
+export CYPRESS_BASE_URL=<value>               # defaults to http://localhost:3000
+export CYPRESS_OC_CLUSTER_USER=<value>        # defaults to jenkins, opt. kubeadmin
```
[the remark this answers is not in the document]
ScriptingShrimp: leaving default to `jenkins` for CI implementation, fixed comments

plugin/cypress.config.ts
ScriptingShrimp: I was trying to set safe default not to interfere with other environment variables I could think of, like `OC_USERNAME` for `oc` cli tool or so. I would say we can definitely something more reasonable like `OC_CONSOLE_USER` and `OC_CONSOLE_PASS` or similar. If we merge this, please feel free to raise this as an issue or refactoring improvement

ScriptingShrimp: variable name changed

plugin/cypress.config.ts
ScriptingShrimp: fixed

plugin/cypress/support/commands.ts
ScriptingShrimp: please see reply in https://github.com/kiali/openshift-servicemesh-plugin/pull/230#discussion_r1445106220, I'm aware for room for improvement.

ScriptingShrimp:
I  was able to run the regression using `yarn cypress:open`, OCP login works for me

ferhoyos (approved):
LGTM. Extra logic to support local CRC can be added in future PR.

PR #232, opened 2023-12-22
Set local Openshift Console version to match OCP version

ferhoyos:
Currently, Openshift console deployed locally (via `start-console.sh`) gets the latest version from quay.io. The issue is that the latest version of Openshift console (4.16) loads Patternfly 5, which is incompatible with last OCP released (4.14) that loads Patternfly 4.

To solve the issue I have calculated OCP server version installed in the local development environment and loads the corresponding Openshift Console version.

### DIFF
--- a/plugin/start-console.sh
+++ b/plugin/start-console.sh
@@ -2,7 +2,8 @@
 
 set -euo pipefail
 
-CONSOLE_IMAGE=${CONSOLE_IMAGE:="quay.io/openshift/origin-console:latest"}
+OPENSHIFT_VERSON=$(oc version | grep "Server Version: " | awk '{print $3}' | cut -d. -f-2)
+CONSOLE_IMAGE=${CONSOLE_IMAGE:="quay.io/openshift/origin-console:$OPENSHIFT_VERSON"}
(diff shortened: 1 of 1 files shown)


jmazzitelli (approved):

plugin/start-console.sh
```diff
@@ -2,7 +2,8 @@
 
 set -euo pipefail
 
-CONSOLE_IMAGE=${CONSOLE_IMAGE:="quay.io/openshift/origin-console:latest"}
+OPENSHIFT_VERSON=$(oc version | grep "Server Version: " | awk '{print $3}' | cut -d. -f-2)
```
jmazzitelli, line 5: just FYI: this assumes (and will only work) if you are already logged into an OpenShift cluster. Otherwise, "oc version" won't tell you about a "Server Version" :
```
$ oc version
Client Version: 4.13.0-202305312300.p0.g05d83ef.assembly.stream-05d83ef
Kustomize Version: v4.5.7
```

and the rest of this script will fail somewhere along the way.

I would suggest doing something like this at the top on line 4: https://github.com/kiali/kiali/blob/master/Makefile#L228

PR #233, opened 2024-01-02
Enable OSSMC plugin if no other plugins are installed in Openshift

ferhoyos:
If no other plugins are installed in Openshift, command `make enable-plugin` fails because `plugin` field is not specified.

This PR first creates `plugin` field if not present. Then add `ossmconsole` entry to `plugin` field, showing a console message if the plugin is already enabled.


jmazzitelli (approved):
tested, lgtm

PR #235, opened 2024-01-09
Ensure oc login before starting plugin console

ferhoyos:
Suggested by @jmazzitelli in https://github.com/kiali/openshift-servicemesh-plugin/pull/232#discussion_r1445131165,  `yarn start-console` comand now throws an error if developer has not login into any OCP.

### DIFF
--- a/plugin/start-console.sh
+++ b/plugin/start-console.sh
@@ -2,6 +2,10 @@
 
 set -euo pipefail
 
+cd ..
+make .ensure-oc-login
+
+cd plugin
 OPENSHIFT_VERSON=$(oc version | grep "Server Version: " | awk '{print $3}' | cut -d. -f-2)
 CONSOLE_IMAGE=${CONSOLE_IMAGE:="quay.io/openshift/origin-console:$OPENSHIFT_VERSON"}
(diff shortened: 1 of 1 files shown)


plugin/start-console.sh
```diff
@@ -2,6 +2,10 @@
 
 set -euo pipefail
 
+cd ..
+make .ensure-oc-login
+
+cd plugin
```
jmazzitelli, line 8: an alternative to the multiple "cd" commands to move around, you can just use the `-C` / `--directory` option to `make`:
```suggestion
make --directory .. .ensure-oc-login
```

jmazzitelli: Or, if you don't want to require "make" in the path, you can just do the same check the makefile does. something like:

```
if ! oc whoami &> /dev/null; then
  echo "You are not logged into an OpenShift cluster. Run 'oc login' before continuing."
  exit 1
fi
```

ferhoyos:
Thanks for the suggestion @jmazzitelli! This change makes code simpler. If it is already defined as `make` command, I prefer to use it instead of copying same code. Besides, this command also checks that `oc` command exists.

jmazzitelli (approved):

PR #245, opened 2024-01-29
Install bookinfo application before graph test [OSSM-5723]

ferhoyos:
To load some known graph and perform more precise in cypress tests, `bookinfo` demo application is installed in Openshift before graph test is run. This test has been modified to check that `bookinfo` graph summary has been loaded correctly.


plugin/cypress/support/commands.ts
```diff
@@ -10,25 +10,25 @@
...
-        cy.get('#inputPassword').type(password)
-        cy.get('button[type="submit"]').click();
-    })
-})
+  cy.visit('/').then(() => {
+    cy.log('OC_IDP: ', typeof idp, JSON.stringify(idp));
+    if (idp != undefined) {
+      cy.get('.pf-c-button').contains(idp).click();
+    }
+    cy.get('#inputUsername').clear().type(user);
```
ferhoyos, line 30: To run cypress test in local CRC I have to clear login input before typing. This should not affect login for OCP instances.

jmazzitelli:
I'm not crazy about duplicating all these hack scripts. We now have dups for download-istio, install-bookinfo-demo, wait-for-namespace .... not sure what others.

We change these hacks often, now we have to worry about changing them in multiple duplicate scripts.

Can we figure out a way to re-use? Even if its just "assume kiali/kiali is git cloned in a peer directory of kiali/openshift-servicemesh-plugin" so we can assume paths like "../../hack/" will work.

ferhoyos:
@jmazzitelli yes I have thought on this... I have just duplicated the files because it is a bit urgent to complete OSSMC GA tests and answer what to do with duplicated files in other PR. Since they are 1:1 copies, maybe it is a good option to duplicate all hacks script from kiali/kiali into `hacks/kiali` folder (similar like we do with application code). What do you think? Is it feasible to extend our `copy-frontend-src-to-ossmc.sh` script to copy hack scripts too?

hack/istio/install-bookinfo-demo.sh
nrfox, line 1: Do we need to copy these scripts directly? If they change in `kiali/kiali` we'll need to remember to update them here as well. Can we reference them as a url instead to keep `kiali/kiali` the source of truth?

nrfox (changes requested):

hack/istio/install-bookinfo-demo.sh
nrfox: Oops looks like mazz commented the same thing as me already...

nrfox: You could do something like `curl -sSL https://github.com/kiali/kiali/raw/master/hack/istio/install-bookinfo-demo.sh | bash`?

ferhoyos: But `install-bookinfo-demo.sh` makes calls to other hack scripts. Will it work if you only download that script?

ferhoyos: Notice that these scripts must be executed for OSSMC tests in Jenkins pipeline, so I am not sure if kiali/kiali git cloned in a peer directory will work for that use case....

jmazzitelli: Or... do a "git pull" or something like that to pull the hack/ directory down using https.

ferhoyos:
I download just `hack/istio` folder from kiali repo using git sparse checkout method. @nrfox @jmazzitelli are you ok with this approach?

jmazzitelli (commented):
Question: Is there a  reason you were not able to put the hack scripts under `_output`? I ask because that's the typical pattern in our kiali repos (the git-ignored and generated output is stored under the main `_output` directory). Is there a need to get these to go under the root "kiali" directory?

e.g. why not put them under `_output/kiali` and then call the scripts directly from there (`_output/kiali/hack/istio/download-istio.sh`)

As it is now, there is going to be this `kiali` directory that will look like its source but isn't really (at least it is git ignored so we won't accidentally check it in, but it does seem out of place).

ferhoyos:
No, there is no reason to be under `kiali` folder. Your approach makes sense, I have put the Kiali scripts under `_output` folder.

jmazzitelli (dismissed):
I haven't run this, but it looks OK to me.

plugin/cypress/support/step_definitions/hooks.ts
```diff
@@ -0,0 +1,127 @@
+import { Before } from '@badeball/cypress-cucumber-preprocessor';
+
+const CLUSTER1_CONTEXT = Cypress.env('CLUSTER1_CONTEXT');
+const CLUSTER2_CONTEXT = Cypress.env('CLUSTER2_CONTEXT');
```
ScriptingShrimp, line 4: Tests don't pass for me locally, bookinfo is not installed, will debug this for more detail.

I think this copypasta might be the issue, I will take a look if we can execute the bookinfo scripts more simpler than utilizing this hook from regular regression - that was maybe a bit over engineered approach 
```
   1) Graph page is displayed correctly
    ✓ Istio Config page is displayed correctly (6999ms)


  3 passing (1m)
  1 failing

  1) Kiali sidebar integration with OCP Console
       Graph page is displayed correctly:
     CypressError: `cy.exec('../_output/kiali/hack/istio/wait-for-namespace.sh -n bookinfo')` failed because the command exited with a non-zero code.

Pass `{failOnNonZeroExit: false}` to ignore exit code failures.

Information about the failure:
Code: 1

Stderr:
Error from server (NotFound): namespaces "kiali-operator" not found

https://on.cypress.io/exec
      at <unknown> (https://console-openshift-console.apps.pmarekipi.maistra.upshift.redhat.com/__cypress/runner/cypress_runner.js:119293:77)
      at tryCatcher (https://console-openshift-console.apps.pmarekipi.maistra.upshift.redhat.com/__cypress/runner/cypress_runner.js:1807:23)
      at Promise._settlePromiseFromHandler (https://console-openshift-console.apps.pmarekipi.maistra.upshift.redhat.com/__cypress/runner/cypress_runner.js:1519:31)
      at Promise._settlePromise (https://console-openshift-console.apps.pmarekipi.maistra.upshift.redhat.com/__cypress/runner/cypress_runner.js:1576:18)
      at Promise._settlePromise0 (https://console-openshift-console.apps.pmarekipi.maistra.upshift.redhat.com/__cypress/runner/cypress_runner.js:1621:10)
      at Promise._settlePromises (https://console-openshift-console.apps.pmarekipi.maistra.upshift.redhat.com/__cypress/runner/cypress_runner.js:1701:18)
      at _drainQueueStep (https://console-openshift-console.apps.pmarekipi.maistra.upshift.redhat.com/__cypress/runner/cypress_runner.js:2407:12)
      at _drainQueue (https://console-openshift-console.apps.pmarekipi.maistra.upshift.redhat.com/__cypress/runner/cypress_runner.js:2400:9)
      at Async._drainQueues (https://console-openshift-console.apps.pmarekipi.maistra.upshift.redhat.com/__cypress/runner/cypress_runner.js:2416:5)
      at Async.drainQueues (https://console-openshift-console.apps.pmarekipi.maistra.upshift.redhat.com/__cypress/runner/cypress_runner.js:2286:14)
  From Your Spec Code:
      at Context.eval (https://console-openshift-console.apps.pmarekipi.maistra.upshift.redhat.com/__cypress/tests?p=cypress/e2e/featureFiles/sidebar-navigation.feature:15941:34)

```

ScriptingShrimp (changes requested):

plugin/cypress/support/step_definitions/hooks.ts
ferhoyos: I have simplified the hook file removing tags that are not used in OSSMC tests and fixed a small issue in OSSMC creation yaml in manage-ossmc.sh script. Anyway these changes will not fix your issue. How have you installed OSSMC? Error says `kiali-operator` namespace is not found, you need the kiali operator installed to install OSSMC right?

ScriptingShrimp: I do install OSSMC by executing Cypress, which do install OSSMC [via this script ](https://github.com/kiali/openshift-servicemesh-plugin/blob/f8e93a36a41d6d06579057398256e41392de62a5/hack/manage-ossmc.sh#L11). The yaml is copied from [Kiali website](https://kiali.io/docs/installation/installation-guide/creating-updating-ossmconsole-cr/#installing-via-oc-cli) how to install OSSMC, so if I understand this correctly the Kiali operator belongs into `openshift-operators` namespace. 

I tried to modify scripts to execute the cypress hooks in correct namespace, still no luck. Currently I'm provisioning fresh cluster to verify whole downstream pipeline - its possible I have something miss-configured in my environment

ferhoyos: Yes, you are right. If you install Kiali operator from Openshift OperatorHub the Kiali operator belongs to `openshift-operators` namespace, not the `kiali-operator` one. In that case we will have to create specific `wait-for-namespace` script for OSSMC or modify the current one to include this case.

ScriptingShrimp: I retested whole pipeline and I again see installed Kiali operator like: 
image

So there is deployment named `kiali-operator` which do belong into `openshift-operators` namespace

The `wait-for-namespace` and other hacks scripts works for normal regression, so I'm not exactly sure where is the difference, imho we should look into that before modyfing hack scripts?

jmazzitelli:
I haven't read all of the notes here, but here are some things I think that will be helpful.

First, do not hardcode the namespace where you think the CR will be. It could be in `kiali-operator` if using our make targets to install things; or it could be in `openshift-operators` if things are installed via OLM. What you want your script to do is to find it dynamically. This is how you do it (I will assume you have one and only one Kiali CR installed):

```
KIALI_CR_NAMESPACE="$(oc get kiali --all-namespaces -o jsonpath='{..metadata.namespace}')"
```

That will output the namespace where the Kiali CR is located. If you do not know the name of the Kiali CR either, then you can do this to get the name and the namespace of the Kiali CR:

```
KIALI_CR_NAMESPACE_NAME="$(oc get kiali --all-namespaces -o custom-columns=NS:.metadata.namespace,N:.metadata.name --no-headers | sed 's/  */:/g')"
KIALI_CR_NAMESPACE="$(echo ${KIALI_CR_NAMESPACE_NAME} | cut -d: -f1)"
KIALI_CR_NAME="$(echo ${KIALI_CR_NAMESPACE_NAME} | cut -d: -f2)"
```

OK, at this point, you have the Kiali CR namespace (and I assume you know the name, too). Next you want to wait for the operator to finish reconciling the CR. We do something similar in our molecule tests as you see [here](https://github.com/kiali/kiali-operator/blob/v1.79.0/molecule/common/wait_for_kiali_cr_changes.yml#L12-L13)):

1. Get the reason and status for condition whose message is "Awaiting next reconciliation" (if there IS no condition whose message is "Awaiting next reconciliation" then the reconciliation is not done yet and you need to keep waiting):
KIALI_CR_REASON="$(oc get kiali $KIALI_CR_NAME -n $KIALI_CR_NAMESPACE -o jsonpath='{.status.conditions[?(@.message=="Awaiting next reconciliation")].reason}')"
KIALI_CR_STATUS="$(oc get kiali $KIALI_CR_NAME -n $KIALI_CR_NAMESPACE -o jsonpath='{.status.conditions[?(@.message=="Awaiting next reconciliation")].status}')"
2. If KIALI_CR_REASON is  "Successful" and KIALI_CR_STATUS is "True" then you can stop waiting - the operator has finished reconciling the CR. Otherwise keep waiting.

I am not sure if you can use those conditions in a "oc wait" command. I thought it was possible, but I can't find it in my notes - I looked into this a while ago. If I find it, I'll add another comment, but you can easily do the above in a script. Something like (I just tested this myself and it works):

```bash
KIALI_CR_NAMESPACE_NAME="$(oc get kiali --all-namespaces -o custom-columns=NS:.metadata.namespace,N:.metadata.name --no-headers | sed 's/  */:/g')"
KIALI_CR_NAMESPACE="$(echo ${KIALI_CR_NAMESPACE_NAME} | cut -d: -f1)"
KIALI_CR_NAME="$(echo ${KIALI_CR_NAMESPACE_NAME} | cut -d: -f2)"

echo -n "Waiting for operator to finish reconciling the CR named [$KIALI_CR_NAME] located in namespace [$KIALI_CR_NAMESPACE]"
while [ "$KIALI_CR_REASON" != "Successful" -o "$KIALI_CR_STATUS" != "True" ]; do
  sleep 1
  echo -n "."
  KIALI_CR_REASON="$(oc get kiali $KIALI_CR_NAME -n $KIALI_CR_NAMESPACE -o jsonpath='{.status.conditions[?(@.message=="Awaiting next reconciliation")].reason}')"
  KIALI_CR_STATUS="$(oc get kiali $KIALI_CR_NAME -n $KIALI_CR_NAMESPACE -o jsonpath='{.status.conditions[?(@.message=="Awaiting next reconciliation")].status}')"
done
echo
echo "Done reconciling"
```

plugin/cypress/support/step_definitions/hooks.ts
jmazzitelli: FYI: by default, the Kiali Server will be installed in the same namespace where the Kiali CR is. If you are running in an OLM environment where the operator is installed in "openshift-operators" and you create the CR via the OpenShift Console GUI, the Kiali CR will be created in "openshift-operators". By default, therefore, the Kiali Server will be installed in that namespace, too. If you do not want that behavior, override the default by explicitly setting `spec.deployment.namespace` in the Kiali CR. For the docs on this setting, see: https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.deployment.namespace
```
.spec.deployment.namespace
The namespace into which Kiali is to be installed. If this is empty or not defined, the default will be the namespace where the Kiali CR is located.
```

hack/wait-for-namespace.sh
```diff
@@ -0,0 +1,74 @@
...
+
+if [ "${IS_MAISTRA}" == "false" ]; then
+  KIALI_CR_NAMESPACE_NAME="$(oc get kiali --all-namespaces -o custom-columns=NS:.metadata.namespace,N:.metadata.name --no-headers | sed 's/  */:/g')"
+  KIALI_CR_NAMESPACE="$(echo ${KIALI_CR_NAMESPACE_NAME} | cut -d: -f1)"
+  KIALI_CR_NAME="$(echo ${KIALI_CR_NAMESPACE_NAME} | cut -d: -f2)"
+  ACCESSIBLE_NAMESPACES="$(oc get kiali $KIALI_CR_NAME -n $KIALI_CR_NAMESPACE -o jsonpath='{.spec.deployment.accessible_namespaces}')"
+
+  # All namespaces are accessible, no need to add namespaces access to CRD
+  if [ "${ACCESSIBLE_NAMESPACES}" != "**" ]; then
+    # accessible_namespaces field is not defined in the CRD, initializing value to empty array
```
jmazzitelli, line 48: I think you mean the CR ("Kiali CR") -- the CRD is the kialis.kiali.io CustomResourceDefinition.

hack/wait-for-namespace.sh
```diff
@@ -0,0 +1,74 @@
...
+      KIALI_CR_STATUS="$(oc get kiali $KIALI_CR_NAME -n $KIALI_CR_NAMESPACE -o jsonpath='{.status.conditions[?(@.message=="Awaiting next reconciliation")].status}')"
+    done
+    echo
+    echo "Done reconciling"
+  fi
+fi
+
+for NAMESPACE in ${NAMESPACES[@]}; do
+  oc wait --for=condition=Ready pods --all -n "$NAMESPACE" --timeout 60s || true
+  oc wait --for=condition=Ready pods --all -n "$NAMESPACE" --timeout 60s
```
jmazzitelli, line 71: Don't understand why wait is done twice here (?).

jmazzitelli (approved):
I didn't run any of this, but I assume it is doing what you need it to do.

hack/wait-for-namespace.sh
ferhoyos: Not sure, this part is copied from `wait-for-namespace.sh` script in Kiali repo, made by @matejnesuta in https://www.github.com/kiali/kiali/pull/6257. Do you remember why wait is done twice here? Probably there is a good reason to do that.

hack/wait-for-namespace.sh
ferhoyos: Oops, yes I mean CR. Typo fixed

hack/wait-for-namespace.sh
matejnesuta: Hello Fernando. The code in the script originated from one of our Jenkins jobs, which is used for installing the demoapps. I had to re-use this piece of code in my Cypress hook, otherwise it would not work on our OCP pipelines, so I decided to create this little script and it is now executed both by the hook and by the Jenkins pipeline. 

Frankly, I am also not sure why the line is present twice here.

ferhoyos: Ok, in that case I will remove the `oc wait` command duplicated

jmazzitelli: I see this code at the end of this script:

```bash
for NAMESPACE in ${NAMESPACES[@]}; do
  oc wait --for=condition=Ready pods --all -n "$NAMESPACE" --timeout 60s
done

sleep 80
```

It looks like this is simply waiting for everything to start (i.e. it waits for all pods in each namespace to be ready, then it waits another 80 seconds before exiting).

Notice the line `oc wait` has a timeout of 60s. If 60 seconds passes and all the pods are not ready, the `oc wait` will exit with a non-zero exit code. This will abort the script (due to `set -e` on line 6, which tells the script to exit on any command that results in an error). That means it will not wait for any more namespaces to be ready and it will not wait that extra 80 seconds.

Now, notice the line you removed in the [last commit](https://github.com/kiali/openshift-servicemesh-plugin/pull/245/commits/8a92d07ebfa45ea085994dbb05a4771714460a9b) had `|| true` at the end - this means even if the `oc wait` timed out, it will not cause the script to abort. So whoever wrote that code originally must have wanted that code to keep waiting even if one namespace isn't ready (i.e. he didn't want the script to abort if `oc wait` timed out) - he wanted to wait for the other namespaces and that extra 80 seconds even if one namespace wasn't ready after 60s. I guess this is to wait as long as possible so the test components have a chance to start fully before the tests begin to run.

I will leave it up to you if you want to maintain that same behavior as before and put back that `|| true` or if you want to leave it out. I'll approve this PR in case you want to leave it as-is and you can merge.

jmazzitelli (dismissed):

hack/wait-for-namespace.sh
ferhoyos: Good catch @jmazzitelli. Eventually this script is used for cypress test configuration, so I prefer to set `|| true` and don't abort the test process if the pods are not ready (to avoid failures if the system is quite slow that day). If they are not ready after the extra 80s wait then cypress tests will fail afterwards. @matejnesuta what do you think?

jmazzitelli (approved):

hack/wait-for-namespace.sh
matejnesuta: I have no problems removing one of the `oc wait` commands as long as it works.

ScriptingShrimp:
testing this on OCP, I cannot see bookinfo in service mesh, nor in Kiali. Execution of the scripts passes, but the test fail on the visual assertion in the dropdown on the image below

image

hack/wait-for-namespace.sh
```diff
@@ -0,0 +1,73 @@
...
+      echo "Unknown argument [$key]. Aborting."
+      exit 1
+      ;;
+  esac
+done
+
+IS_MAISTRA=$([ "$(oc get crd | grep servicemesh | wc -l)" -gt "0" ] && echo "true" || echo "false")
+
+if [ "${IS_MAISTRA}" == "false" ]; then
+  KIALI_CR_NAMESPACE_NAME="$(oc get kiali --all-namespaces -o custom-columns=NS:.metadata.namespace,N:.metadata.name --no-headers | sed 's/  */:/g')"
```
nrfox, line 41: You can ignore this but it's often easier to use jsonpath and/or jq to parse this stuff than it is to chop up the text e.g. `oc get kiali --all-namespaces -o jsonpath='{.items[*].metadata.namespace}'`

hack/wait-for-namespace.sh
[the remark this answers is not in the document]
ferhoyos: Thanks for catching up this @nrfox. I add this script manually because I added some changes here. But I think the best approach is to include these changes into Kiali repo and then download it from there. 

Kiali PR to include script changes: https://github.com/kiali/kiali/pull/7230

ScriptingShrimp:
Hi @ferhoyos , I'm still getting error related to bookinfo on OCP
```
Kiali sidebar integration with OCP Console
    ✓ Service mesh buttons are displayed (17644ms)
    ✓ Overview page is displayed correctly (8283ms)
    1) Graph page is displayed correctly
    ✓ Istio Config page is displayed correctly (9737ms)


  3 passing (8m)
  1 failing

  1) Kiali sidebar integration with OCP Console
       Graph page is displayed correctly:
     CypressError: `cy.exec('../_output/kiali/hack/istio/wait-for-namespace.sh -n bookinfo')` timed out after waiting `400000ms`.
```

were you able to run this locally? 
I'll investigate more in detail tomorrow .

ScriptingShrimp (approved):
I will merge this now, lets see if this passing on CI. 
If not we will revert the merge commit on Monday.

nrfox (approved):

PR #247, opened 2024-02-01
Align release github action with Kiali

### LINKED ISSUE #229: Update the Release Job to be triggered manually
I was trying to release a version but the release job was skipped, see https://github.com/kiali/openshift-servicemesh-plugin/actions/runs/7274246466

It should be possible to, at least be able to specify all the parameters needed to be released, like the Kiali and the Operator.

ferhoyos:
### Describe the change

Now that openshift-servicemesh-plugin follows same release cadence as kiali (one minor release every 3 weeks and branch creation with that release), I have aligned release github action with kiali repository one to determine what type of release needs:

- If the branch selected is `main` -> Minor version (e.g., from 1.79 to 1.80)
- If the branch selected is a release version -> Patch version (e.g. from 1.79.0 to 1.79.1 if branch selected is v1.79

Currently patch versions are not released because OSSMC always embeds 1.73 version due to PF limitation. When the plugin is adapted to PF5 we will have to plan a release schedule to match Kiali version with OSSMC version (I mean Kiali code embedded in OSSMC).

### Issue reference
Closes #229

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -3,8 +3,7 @@ name: Release
 on:
   schedule:
   # Every Monday at 07:00 (UTC)
-  - cron: '00 7 * * MON'
-
+  - cron: "00 7 * * MON"
   workflow_dispatch:
     inputs:
       plugin_quay_repository:
@@ -24,10 +23,9 @@ jobs:
       branch_version: ${{ env.branch_version }}
       next_version: ${{ env.next_version }}
       plugin_quay_tag: ${{ env.plugin_quay_tag }}
-      plugin_quay_repo: ${{ env.plugin_quay_repo }}
     steps:
     - name: Checkout code
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
       with:
         ref: ${{ github.ref_name }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -143,10 +142,9 @@ jobs:
           PLUGIN_QUAY_REPO="${{ github.event.inputs.plugin_quay_repository }}"
         fi
 
-        PLUGIN_QUAY_TAG="$PLUGIN_QUAY_REPO:$RELEASE_VERSION"
+        PLUGIN_QUAY_TAG="$PLUGIN_QUAY_REPO:$RELEASE_VERSION $PLUGIN_QUAY_REPO:$BRANCH_VERSION"
```
josunect, line 145: Just some clarification on this line: It is right to have a tag with blanks?

ferhoyos: Yes, the blank means that you will have two tags (the release version and the branch version, for example v1.78.0 and v1.78). It is exactly what we have in kiali release process (https://github.com/kiali/kiali/blob/24ca2f7ade3222a99664654715286e420a4325a6/.github/workflows/release.yml#L146)

josunect (approved):
LG

PR #248, opened 2024-02-05
exception handling for Cypress

ScriptingShrimp:
copied exception handling (from kiali/kiali) to handle unsuccessful API calls 

I came across this error in Jenkins pipeline , this should fix that.

> Error Message
> The following error originated from your application code, not from Cypress. It was caused by an unhandled promise rejection.    > Request failed with status code 403  When Cypress detects uncaught errors originating from your application it will automatically fail the current test.  This behavior is configurable, and you can choose to turn this off by listening to the `uncaught:exception` event.  https://on.cypress.io/uncaught-exception-from-application
> Stacktrace
> Error: The following error originated from your application code, not from Cypress. It was caused by an unhandled promise rejection.
> 
>   > Request failed with status code 403
> 
> When Cypress detects uncaught errors originating from your application it will automatically fail the current test.
> 
> This behavior is configurable, and you can choose to turn this off by listening to the `uncaught:exception` event.
>

### DIFF
--- a/plugin/cypress/e2e/workloads-deployments.ts
+++ b/plugin/cypress/e2e/workloads-deployments.ts
@@ -1,29 +1,45 @@
-import { Given, Then, When } from "@badeball/cypress-cucumber-preprocessor";
+import { Before, Then, When } from '@badeball/cypress-cucumber-preprocessor';
+
+Before(() => {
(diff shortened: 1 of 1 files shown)


ferhoyos (approved):
LGTM

PR #259, opened 2024-02-12
Load MTLS status and certs info

### LINKED ISSUE #253: TLS information is not available
In the overview page, TLS information is not available:

image

It only affects OSSMC, because that information is available in Kiali server:

image

ferhoyos:
Since OSSMC does not render `MeshMTLSStatus` and `IstioCertsInfo` components located in Kiali header, MLTS and certs info are not loaded.

Correspondent API calls and redux set methods have been aded to `KialiController` class to store this information and make it available through OSSMC plugin.

How to test this PR:
1. Go to Service Mesh > Overview page
2. Check security info:
  - MTLS: Check `Min TLS Version` field of Istio namespace. Some value should appear next to the lock icon
  
image

  - Certs Info: If you hover `Min TLS Version` field of Istio namespace, certification info should appear.

image

Since this bug has high priority, I have created different issue to track Cypress tests of this feature (https://github.com/kiali/openshift-servicemesh-plugin/issues/265)

Fixes #253

### DIFF
--- a/plugin/src/openshift/components/KialiController.tsx
+++ b/plugin/src/openshift/components/KialiController.tsx
@@ -87,7 +87,6 @@ class KialiControllerComponent extends React.Component<KialiControllerProps> {
   private promises = new PromisesRegistry();
 
   state = {
-    namespaces: [],
     configLoaded: false
   };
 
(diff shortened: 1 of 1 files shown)


plugin/src/openshift/components/KialiController.tsx
```diff
@@ -131,16 +142,42 @@ class KialiControllerComponent extends React.Component<KialiControllerProps> {
...
+          }
+        });
+
+      const getIstioCertsInfoPromise = this.promises
+        .register('getIstioCertsInfo', API.getIstioCertsInfo())
+        .then(response => this.props.setIstioCertsInfo(response.data));
+
+      await Promise.all([getMeshMTLSPromise, getIstioCertsInfoPromise]);
+    } catch (err) {
+      console.error('Error loading kiali security info', err);
```
josunect, line 176: Shouldn't this be also added as an AlertUtils to warn the user?

josunect (commented):
Current behavior: 
image

With PR: 

image

plugin/src/openshift/components/KialiController.tsx
ferhoyos: I think it is not worth to show to the user the global error message + specific ones. For me it is enough to show the specific ones to the user. Besides, I have noticed that AlertUtils error messages are not shown in OSSMC. I will create separate issue since this is not part of this PR.

josunect: Well, that was for the ones that were not captured. I think it is better to show the specific error message, the only problem that I see, is if there is a general problem in the network or the access to the Kiali proxy and the user receives too many alerts.

josunect (approved):
LG

plugin/src/openshift/components/KialiController.tsx
ferhoyos: Yes, although on the other hand it shows to the user the feeling that something really wrong is happening to the application hahaha

PR #272, opened 2024-03-04
start release build at 1am

jmazzitelli:
This is a temp fix to https://github.com/kiali/kiali/issues/6865 - read that issue to find out why this is being done.


jmazzitelli:
This is only a suggested fix. I can't think of another way to quickly (albeit temporarily) fix the issue.

jshaughn (approved):
Make it so, #2.

PR #275, opened 2024-03-08
Adapt Cypress tests to OpenShift 4.15

### LINKED ISSUE #274: Cypress tests are not passed in OpenShift 4.15
Following errors are found when cypress tests are executed on OpenShift 4.15:
- In `sidebar-navigation.feature`:
```
AssertionError: Timed out retrying after 40000ms: Expected to find element: `button.pf-c-nav__link`, but never found it.
``` 

- In `workloads-deployments.feature`:
```

```

Probably this is due to the fact that OpenShift 4.15 uses version 5 of Patternfly library (OSSMC uses PF4). Although there exists compatibility backwards to PF4 plugins, it seems that tests are not adapted.

Note that Cypress tests must pass for all supported OpenShift versions (currently from 4.12 until 4.15).

ferhoyos:
### Describe the change
This PR adapts Cypress tests to work in all supported OpenShift versions (from 4.12 to 4.15).

Since OpenShift 4.15 uses version 5 of Patternfly library, all CSS classes have changed from `pf-c` to `pf-v5-c`, so current selectors used in Cypress do not work. To avoid this issue I have used selector [class$=".."] to check only the last part of the CSS class.

Regarding MobX issue, I just ignore uncaught exceptions that contain any `Mobx` reference.

As a bonus I have adapted Cypress tests to be able to run in localhost due to slightly differences with 'standard' OpenShift:
- There is no login page
- API_PROXY value is different

### Issue Reference
Fixes #274

### DIFF
--- a/plugin/cypress/e2e/sidebar-navigation.ts
+++ b/plugin/cypress/e2e/sidebar-navigation.ts
@@ -1,67 +1,94 @@
-import { Given, Then, When } from "@badeball/cypress-cucumber-preprocessor";
+import { Before, Then, When } from '@badeball/cypress-cucumber-preprocessor';
+import { getApiProxy } from '../support/utils';
+
+Before(() => {
+  // This prevents cypress from stopping on errors unrelated to the tests.
+  // There can be random failures due timeouts/loadtime/framework that throw browser errors.  This
+  // prevents a CI failure due something like a "slow".  There may be a better way to handle this.
+  cy.on('uncaught:exception', (err, _runnable, promise) => {
+    // when the exception originated from an unhandled promise
+    // rejection, the promise is provided as a third argument
+    // you can turn off failing the test in this case
+    if (promise || err.message.includes('MobX')) {
+      return false;
+    }
+    // we still want to ensure there are no other unexpected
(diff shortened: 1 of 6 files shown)


plugin/cypress/e2e/sidebar-navigation.ts
```diff
@@ -1,67 +1,94 @@
...
+
+Before(() => {
+  // This prevents cypress from stopping on errors unrelated to the tests.
+  // There can be random failures due timeouts/loadtime/framework that throw browser errors.  This
+  // prevents a CI failure due something like a "slow".  There may be a better way to handle this.
+  cy.on('uncaught:exception', (err, _runnable, promise) => {
+    // when the exception originated from an unhandled promise
+    // rejection, the promise is provided as a third argument
+    // you can turn off failing the test in this case
+    if (promise || err.message.includes('MobX')) {
```
ScriptingShrimp, line 12: I saw the MobX error messages on failed installation of OSSMC, but was unable to reproduce so far

ScriptingShrimp (approved):
passing on OCP, also in jenkins pipeline

PR #276, opened 2024-03-08
Isolate different versions of MobX for OCP 4.15

ferhoyos:
### Describe the change
In OpenShift 4.15, UI consoles uses version 5.1.0 of  `@patternfly/react-topology` library, which uses version 6.10.0 of `mobx` library.

On the other hand, OSSMC uses version 4.91.27 of `@patternfly/react-topology`, which uses version 5.15.7 of `mobx`.

Since multiple `mobx` instances are being used with different versions, following console error appears in the `Graph` page:
```

```

To fix the issue I have set `configure({ isolateGlobalState: true })` to any OSSMC page where graph is shown.

### Steps to test the PR
Go to `Graph` page and check that no console error related to `MobX` appears in the logs.

### Issue reference
#274


ferhoyos:
Closing this PR since it is not valid when the plugin upgrades to PF5

PR #283, opened 2024-03-19
README - how to deploy dev build of plugin via operator

jmazzitelli:


ferhoyos (approved):
This info is quite useful to deploy local plugin in remote OpenShift cluster. Thanks @jmazzitelli!

PR #287, opened 2024-03-26
[v1.73] Isolate different versions of MobX for OCP 4.15

ferhoyos:
### Describe the change
In OpenShift 4.15, the UI console uses the version 5.1.0 of  `@patternfly/react-topology` library, which uses the version 6.10.0 of the `mobx` library.

On the other hand, OSSMC uses the version 4.91.27 of the `@patternfly/react-topology`, which uses the version 5.15.7 of `mobx`.

Since multiple `mobx` instances are being used with different versions, the following console error appears in the `Graph` page:
```

```

To fix the issue I have set `configure({ isolateGlobalState: true })` to any OSSMC page where the graph is shown.

I have also fixed a small style issue in the context menu:

Before:

image

Now:

image

### Steps to test the PR
Go to `Graph` page and check that no console error related to `MobX` appears in the logs.

### Issue reference
#274

### DIFF
--- a/plugin/src/openshift/components/KialiContainer.tsx
+++ b/plugin/src/openshift/components/KialiContainer.tsx
@@ -22,7 +22,13 @@ import 'ace-builds/src-noconflict/ext-searchbox';
 const ossmcStyle = kialiStyle({
   display: 'flex',
   flexDirection: 'column',
-  overflowY: 'auto'
+  overflowY: 'auto',
+  $nest: {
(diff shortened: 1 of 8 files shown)


jshaughn (commented):
This is tricky, good catch.  Again, just wondering about `yarn.lock` change w/o package.json change.

jshaughn (approved):

Issue #289, opened 2024-04-02
Cypress test for Istio config list page


ferhoyos:
### What do you want to improve?

Add Cypress test for the Istio config list page, the only native page for Openshift (it does not import any code from the Kiali repository).

### What is the current behavior?

There is no Cypress test to verify the correct behaviour of the Istio Config List page.

### What is the new behavior?

Existing cypress test to verify the correct behaviour of the Istio config list page in automated E2E tests.

PR #290, opened 2024-04-02
Initialize the load settings of the Istio config list page when switching namespaces

### LINKED ISSUE #288: The Istio config list page does not update when switching from a forbidden namespace to an accessible one
### Describe the bug

When navigating to the Istio config list page and switching to a namespace that is not accessible (configured in Kiali server config), a `Forbidden` error message is displayed. The issue arises when the config list page fails to update upon switching to an accessible namespace, leaving the user stuck with the `Forbidden` message.

image

### Expected Behavior
When switching from a forbidden namespace to an accessible one, the Istio config list should display items from the accessible namespace.

### What are the steps to reproduce this bug?

1. Change Kiali server config to allow only specific namespace to be accessible (avoid using the `**` value).
2. Navigate to Service Mesh -> Istio Config List.
3. Switch to forbidden namespace and verify that the `Forbidden` message appears.
4. Switch to accessible namespace and verify that the `Forbidden` message of previous namespace is still there.

### Environment

- **OSSMC/Kiali version:** 1.73
- **OpenShift version:** 1.14
- **Istio version:** 1.18

ferhoyos:
### Describe the change

Initialize the load settings of the Istio config list page when switching namespaces.

### Steps to test the PR

1. Change Kiali server config to allow only specific namespace to be accessible (avoid using the `**` value).
2. Navigate to Service Mesh -> Istio Config List.
3. Switch to forbidden namespace and verify that the `Forbidden` message appears.
4. Switch to accessible namespace and verify that the Istio Config items list is displayed.

### Automation testing

Automation testing for the Istio config list page to be implemented (https://github.com/kiali/openshift-servicemesh-plugin/issues/289)

### Issue reference

Fixes #288

### DIFF
--- a/plugin/src/openshift/pages/IstioConfigListPage.tsx
+++ b/plugin/src/openshift/pages/IstioConfigListPage.tsx
@@ -165,8 +165,8 @@ const newIstioResourceList = {
 const IstioConfigListPage = () => {
   const { ns } = useParams<{ ns: string }>();
   const [loaded, setLoaded] = React.useState<boolean>(false);
-  const [listItems, setListItems] = React.useState<any[]>([]);
-  const [loadError, setLoadError] = React.useState<OSSMCError>();
+  const [listItems, setListItems] = React.useState<IstioConfigObject[]>([]);
+  const [loadError, setLoadError] = React.useState<OSSMCError | null>(null);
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
LG!

PR #292, opened 2024-04-04
fix release workflow

jmazzitelli:
get this to be identical to main branch. we need to configure git here otherwise the build fails.


jshaughn (approved):
OK

Issue #298, opened 2024-04-11
Istio config list page does not filter by namespace (OCP 4.15)


ferhoyos:
### Describe the bug

The istio config list page does not filter the istio items list by namespace when the user selects any of the namespaces in the filter component.

![image](https://github.com/kiali/openshift-servicemesh-plugin/assets/122779323/93e89a3a-6de6-4ce3-b82a-00afdba1b2d4)

*Note:* This bug only happens for OpenShift 4.15 version

### Expected Behavior

The istio config list should be filtered by namespace when the user selects any of the namespaces in the filter component.

### What are the steps to reproduce this bug?

1. Navigate to the istio config list page
2. Click on the namespace filter component
3. Select any of the namespaces
4. Verify that istio config list items is not filtered by namespace

### Environment

- **OSSMC/Kiali version:** 1.73
- **OpenShift version:** 4.15
- **Istio version:** 1.18

PR #299, opened 2024-04-11
[v1.73] Fix Istio config list filter by namespace

ferhoyos:
### Describe the change

Fix the Istio config list namespace filter. This bug only affects OCP 4.15

### Steps to test the PR

1. Navigate to the istio config list page
2. Click on the namespace filter component
3. Select any of the namespaces
4. Verify that the istio config items are filtered by namespace

### Issue reference

#298


ferhoyos:
Don't merge this PR since it breaks support for OCP < 4.15

ferhoyos:
Closing since there is a workaround using `useActiveNamespace` instead of  `useParams` that add support for OpenShift 4.14.

Issue #304, opened 2024-04-19
'Node Graph' context menu option does not work in OCP 4.15


ferhoyos:
### Describe the bug

In OpenShift 4.15, when the user clicks on the 'Node graph' option of the traffic graph context menu of any node, the node graph details are not displayed. In spite of this, the page is refreshed with the same traffic graph.

### Expected Behavior

The console should show the node graph details when the user clicks on the 'Node graph' option of the traffic graph context of any node.

### What are the steps to reproduce this bug?

1. Go to the Graph page
2. Click on the context menu of any node
3. Select the option 'Node Graph'
4. Verify that the console does not display the node graph details. Instead, the global traffic graph is refreshed.

![image](https://github.com/kiali/openshift-servicemesh-plugin/assets/122779323/ae7d30cd-4235-43e4-9acc-a31bd49e9dff)

### Environment

- **OSSMC/Kiali version:** 1.73
- **OpenShift version:** 4.15
- **Istio version:** 1.18

Issue #254, opened 2024-02-09
Add Cypress tests to check OSSMC link redirections


ferhoyos:
Kiali has a different internal path structure than OSSMC (e.g., urls for Kubernetes items like services or workloads). For this reason Kiali internal link events are captured by OSSMC listener and adapted to OSSMC link structure. A cypress test should be created to test this feature.

ferhoyos:
The different internal path structure of OSSMC is handled by overwriting the visit command in the Kiali Cypress tests that are executed in OSSMC:

https://github.com/kiali/openshift-servicemesh-plugin/blob/26a847e6d73a32265b4e22eb9458fdba592e3acb/plugin/cypress/support/commands.ts#L142

If there are any issues with OSSMC internal link redirections, the adapted Kiali Cypress tests will detect them. Therefore, there is no need to create a specific OSSMC Cypress test for this.

PR #305, opened 2024-04-22
[v1.73] Obtain the graph params from url pathname

ferhoyos:
### Describe the change

Since `useParams` does not work for OCP 4.15, I have found a workaround to obtain the graph params from url pathname with `useLocation`. This PR allows the user to see the node graph details of a specific node.

### Steps to test the PR

1. Go to the Graph page
2. Click on the context menu of any node
3. Select the option 'Node Graph'
4. Verify that the console does not display the node graph details. Instead, the global traffic graph is refreshed.

image

### Automation testing

To be added in https://github.com/kiali/openshift-servicemesh-plugin/issues/254

### Issue reference

#304


jshaughn (approved):
Annoying that we need to do this, but LGTM!

PR #307, opened 2024-05-06
copy the kiali source prior to building release

jmazzitelli:
This will copy the source code from a kiali server release into the OSSMC plugin source so OSSMC release will be done with the source code for a released kiali server. The default version of kiali source code will be the version of the OSSMC release being built.

fixes: https://github.com/kiali/kiali/issues/6865

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,8 +2,8 @@ name: Release
 
 on:
   schedule:
-  # Every Monday at 01:00 (UTC) - this action takes several hours to complete
-  - cron: "00 1 * * MON"
+  # Every Tuesday at 01:00 (UTC) which is the day after the server release is done. This action takes several hours to complete.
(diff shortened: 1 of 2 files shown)


jmazzitelli:
This is what the release PR will look like (I ran this action on my fork): https://github.com/jmazzitelli/openshift-servicemesh-plugin/pull/5

The image was pushed to quay correctly. I ran the image and it was the correct version that I was expecting.

jshaughn (commented):
Looks good, thanks for taking care of this pretty-quick(ly).

jshaughn (approved):

ferhoyos:
Thank you @jmazzitelli for taking care of this! A very clean PR, the idea of using the same git push for both kiali source code and release version is brilliant. At the beginning I thought that we had to do complicated PR auto-merges and things like that.

PR #309, opened 2024-05-20
Add Mesh page to OSSMC

### LINKED ISSUE #308: Include Mesh page in OSSMC
### What do you want to improve?

Kiali has recently replaced old mesh page with new infrastructure topology graph:

image

This feature is a great improvement in mesh management and it should be included in the OSSMC plugin

### What is the current behavior?

OSSMC does not display the mesh page.

### What is the new behavior?

New `Mesh` navigation entry in the Service Mesh section that displays the new mesh page.

ferhoyos:
### Describe the change

Add Mesh page to OSSMC

**Bonus**: Extract console plugin metadata into separate file, as indicated here: https://github.com/openshift/console/blob/master/frontend/packages/console-dynamic-plugin-sdk/README.md#plugin-metadata

### Steps to test the PR

1. Click to `Service Mesh` / `Mesh` navigation menu
2. Verify that Kiali mesh page is loaded correctly

### Automation testing

Cypress test to check that mesh page has been loaded correctly (detailed tests are included in Kiali repo)

### Issue reference

Closes #308


plugin/plugin-metadata.ts
```diff
@@ -0,0 +1,24 @@
+import { ConsolePluginBuildMetadata } from '@openshift-console/dynamic-plugin-sdk-webpack';
+
+const metadata: ConsolePluginBuildMetadata = {
+  name: 'ossmconsole',
+  version: '1.85.0',
```
jmazzitelli, line 5: During our build process, we have update this version string. Because it is now in a new file, we should make sure our build process doesn't break. I'm not sure exactly where we do this - it is either in the github workflow or in some Makefile. I will look to try to find it.

jmazzitelli: This looks like the place: https://github.com/kiali/openshift-servicemesh-plugin/blob/main/.github/workflows/release.yaml#L209

jmazzitelli: Note that we still need to change "version:" here in package.json (there is one at the top of the file)... but now we need to do this version update in this plugihn-metadata.ts file, too.

jmazzitelli: I think we want SOMETHING this:

```
        jq -r ".version |= \"${RELEASE_VERSION:1}\" plugin/package.json > plugin/package.json.tmp
        mv plugin/package.json.tmp plugin/package.json

        jq -r ".version |= \"${RELEASE_VERSION:1}\" plugin/plugin-metadata.json > plugin/plugin-metadata.json.tmp
        mv plugin/plugin-metadata.json.tmp plugin/plugin-metadata.json
```

But that probably isn't 100% correct because now the plugin version is inside a javascript file, not a json file so the "jq" tool likely will not work. So for that second set of commands we will need to use something like "sed" rather than "jq".

jmazzitelli: see https://github.com/ferhoyos/openshift-servicemesh-plugin/pull/6

jmazzitelli (dismissed):
I created a PR to this PR so it updates the version strings properly in all the files where version string is now found (there is now a third file - plugin-metadata.ts - where version has to be updated by the release action).

See https://github.com/ferhoyos/openshift-servicemesh-plugin/pull/6

You can test that the new script works by running it and checking the three files are correctly changed.

Try it with a snapshot version and non-snapshot version (the package.json and plugin-metadata.ts should not have `-SNAPSHOT` in the version string - the script will strip it out.  The Makefile will still show -SNAPSHOT in the version string).

1. `hack/update-version-string.sh v1.2.3` should result in:
```diff
diff --git a/Makefile b/Makefile
-VERSION ?= v1.85.0-SNAPSHOT
+VERSION ?= v1.2.3

diff --git a/plugin/package.json b/plugin/package.json
-  "version": "1.85.0",
+  "version": "1.2.3",

diff --git a/plugin/plugin-metadata.ts b/plugin/plugin-metadata.ts
-  version: '1.85.0',
+  version: '1.2.3',
```

2. `hack/update-version-string.sh v1.2.3-SNAPSHOT` should result in:
```diff
diff --git a/Makefile b/Makefile
-VERSION ?= v1.85.0-SNAPSHOT
+VERSION ?= v1.2.3-SNAPSHOT

diff --git a/plugin/package.json b/plugin/package.json
-  "version": "1.85.0",
+  "version": "1.2.3",

diff --git a/plugin/plugin-metadata.ts b/plugin/plugin-metadata.ts
-  version: '1.85.0',
+  version: '1.2.3',
```

jmazzitelli (approved):
I can see the new Mesh option:
image

image

.github/workflows/release.yaml
```diff
@@ -255,10 +250,7 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       if: ${{ needs.initialize.outputs.release_type == 'minor' }}
       run: |
-        sed -i -r "s/^VERSION \?= (.*)/VERSION \?= $NEXT_VERSION-SNAPSHOT/" Makefile
-
-        jq -r ".version |= \"${NEXT_VERSION:1}\" | .consolePlugin.version |= \"${NEXT_VERSION:1}\"" plugin/package.json > plugin/package.json.tmp
-        mv plugin/package.json.tmp plugin/package.json
+        hack/update-version-string.sh "${NEXT_VERSION}-SNAPSHOT"
 
         git add Makefile plugin/package.json
```
jmazzitelli, line 255: There is a new file that gets updated when we do a release. We have to make sure we add that new file to the release's git commit:

```suggestion
        git add Makefile plugin/package.json plugin/plugin-metadata.ts
```

jmazzitelli: GitHub won't let me add a comment/suggestion to the other line where this change needs to happen, so I need to just document it here. On line 232, add `plugin/plugin-metadata.ts` to the `git add` command:

```
        git add Makefile plugin/package.json
```

should be:
```
        git add Makefile plugin/package.json plugin/plugin-metadata.ts
```

jmazzitelli (changes requested):
more changes needed for the release action so the new plugin-metadata.ts file gets updated in the release PRs that get generated.

jmazzitelli (approved):

PR #310, opened 2024-05-21
Move base image to ubi9 nginx 1.24

aljesusg:
### Describe the change

Replace base image ubi8/nginx1.20 to ubi9/nginx1.24

### Issue reference

Grade B security in ubi8/nginx1.20


leandroberetta (approved):
Looks good.

Issue #312, opened 2024-05-28
release builds fail - IBM arches are unsupported in the github CI infrastructure


jmazzitelli:
Release builds are now failing.

```
#40 [linux/ppc64le stage-1 3/4] RUN chown -R 1001:0 /usr/share/nginx/html/
#40 0.045 Fatal glibc error: CPU lacks float128 support (POWER 9 or later required)
#40 ERROR: process "/bin/sh -c chown -R 1001:0 /usr/share/nginx/html/" did not complete successfully: exit code: 127
```

```
#32 [linux/s390x stage-1 3/4] RUN chown -R 1001:0 /usr/share/nginx/html/
#32 0.046 Fatal glibc error: CPU lacks VXE support (z14 or later required)
#32 ERROR: process "/bin/sh -c chown -R 1001:0 /usr/share/nginx/html/" did not complete successfully: exit code: 127
```

We currently cannot release OSSMC to the community, unless we turn off ppc64le and s390x architecture support.

jmazzitelli:
Just ran a test with latest buildkit and binfmt and things worked:
```
#31 [linux/s390x stage-1 3/4] RUN chown -R 1001:0 /usr/share/nginx/html/
#31 DONE 0.5s
#32 [linux/ppc64le stage-1 3/4] RUN chown -R 1001:0 /usr/share/nginx/html/
#32 DONE 0.6s
```
So we need to upgrade this: https://github.com/kiali/openshift-servicemesh-plugin/blob/v1.85.0/make/Makefile.plugin.mk#L133-L144

jmazzitelli:
I got it to work with this:

```diff
diff --git a/make/Makefile.plugin.mk b/make/Makefile.plugin.mk
index e25798a..1226220 100644
--- a/make/Makefile.plugin.mk
+++ b/make/Makefile.plugin.mk
@@ -134,14 +134,14 @@ prepare-dev-env: .determine-kiali-url
 .ensure-buildx-builder: .ensure-docker-buildx
        @if ! docker buildx inspect ossmconsole-builder > /dev/null 2>&1; then \
          echo "The buildx builder instance named 'ossmconsole-builder' does not exist. Creating one now."; \
-         if ! docker buildx create --name=ossmconsole-builder --driver-opt=image=moby/buildkit:v0.12.5 --config ./make/buildkitd.toml --use; then \
[...]

PR #313, opened 2024-05-28
do not build ppc64le arch for now

jmazzitelli:
temporarily works around https://github.com/kiali/openshift-servicemesh-plugin/issues/312

We will want to revert this PR once we figure out how to get ppc64le built again.


jshaughn (approved):
+1, we can nix problematic upstream platforms.

jmazzitelli:
FWIW: I was able to build PPC64LE on my local machine. If all else fails , we can always have someone build a release locally and push that. But for now, let's just skip it and maybe we can figure out how to get github to build this.

PR #319, opened 2024-06-18
[hack] do not consider it an error if nothing has changed

jmazzitelli:
If the main branch already has the updated Kiali server repo code, the hack script that is run by the release action will attempt to add/commit zero changed files and that results in "git commit" returning a non-zero exit code. This aborts the hack script and the release action assumes a failure occurred.

This is actually OK to happen - it just means we already updated the main branch with the latest code.

This PR checks to see if nothing needs to be committed, and if so, it just exits normally since this is an OK condition.

### DIFF
--- a/hack/copy-frontend-src-to-ossmc.sh
+++ b/hack/copy-frontend-src-to-ossmc.sh
@@ -165,17 +165,22 @@ ${COMMIT_MESSAGE}
 EOM
 
 git add ${ABS_DEST_DIR}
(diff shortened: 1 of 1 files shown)


jmazzitelli:
The `--quiet` option to `git diff-index` makes the command exit with a status of 1 if there are differences (files to be committed) and 0 if there are none, without showing any output. It implies the `--exit-code` option. See `git diff-index --help` for details.

PR #320, opened 2024-06-18
[github action] Copy Kiali source first before updating version numbers.

jmazzitelli:
Rather than first updating the Makefile and related files with the new version, we need to copy the Kiali source code first. This is so we can avoid errors when nothing needs to be copied from the Kiali server repo.


jmazzitelli:
I'm running a test on my fork to make sure this fixes everything. I'll wait for that to succeed before merging this.

jmazzitelli:
my test succeeds.. going to merge this and do a release now

PR #324, opened 2024-06-28
Align PF libraries with Kiali

ferhoyos:
### Describe the change

Due to a bug in the OpenShift console, the plugin could not use PF version 5.2+ and we had to downgrade to 5.1.x. Now that the issue is fixed in version `1.0.2` of the library `@openshift-console/dynamic-plugin-sdk-webpack`  (https://www.github.com/openshift/console/pull/13832), we can align the PF library with the version used in the Kiali standalone application. Other libraries that were recently updated for Kiali v1.86 have also been updated.

Additionally, I have modified some CSS to have a gray background on Service and Workload details:
Before:

image

After:

image

### Steps to test the PR

OSSMC smoke test to verify that the PF upgrade does not break anything.

### Automation testing

N/A


plugin/package.json
```diff
@@ -28,14 +28,13 @@
...
   "dependencies": {
-    "@openshift-console/dynamic-plugin-sdk": "^1.3.0",
-    "@patternfly/patternfly": "5.1.0",
-    "@patternfly/react-charts": "7.1.1",
-    "@patternfly/react-core": "5.1.1",
-    "@patternfly/react-icons": "5.1.1",
-    "@patternfly/react-styles": "5.1.1",
-    "@patternfly/react-table": "5.1.1",
-    "@patternfly/react-topology": "5.2.1",
+    "@openshift-console/dynamic-plugin-sdk": "1.0.0",
```
ferhoyos, line 31: According to https://github.com/openshift/console/blob/master/frontend/packages/console-dynamic-plugin-sdk/README.md#openshift-console-versions-vs-sdk-versions, the version fully compatible with Openshift Console 4.15 is the version `1.0.0`. For this reason I prefer to downgrade to avoid unexpected behaviours.

ferhoyos:
That makes sense @jshaughn. If the new release is fairly soon, I have no problem to wait for 5.4.0 and align both Kiali and OSSMC with that version.

jshaughn:
Actually, it may be a few weeks.  We can merge this now and update PFT independently when the release comes out.

PR #327, opened 2024-07-01
Service mesh tab is loaded correctly after refresh

ferhoyos:
### Describe the change

If the user goes to a Service details page, clicks on the `Service Mesh` tab, and refreshes the page, the Service Mesh details page gets stuck on the service mesh content info even though the active tab is the default `Details` tab. This PR fixes the issue by properly setting the router basename.

### Steps to test the PR

1. In OpenShift 4.15, go to any service details page
2. Click on the `Service Mesh` tab
3. Refresh the page
4. Verify that the service mesh tab content is shown and the active tab is the `Service Mesh` one.

image

5. Verify that you can navigate correctly to the rest of the tabs of the Service Mesh.

image

### Automation testing

To avoid conflicts due to test refactorization in #323, a test scenario covering this issue will be added in that PR or in a separate PR.

### Issue reference

Fixes #325

PR #328, opened 2024-07-01
[v1.73] Service mesh tab is loaded correctly after refresh

ferhoyos:
Backport of PR #327 for branch v1.73

I have also fixed a small issue with `New Create` option in the Istio config list page with `all-namespaces` value.

### DIFF
--- a/plugin/src/openshift/utils/KialiIntegration.ts
+++ b/plugin/src/openshift/utils/KialiIntegration.ts
@@ -2,6 +2,7 @@ import { consoleFetchJSON } from '@openshift-console/dynamic-plugin-sdk';
 import { useHistory } from 'react-router-dom';
 import { refForKialiIstio } from './IstioResources';
 import { History } from 'history';
+import { setHistory } from 'app/History';
 
 export const properties = {
   // This API is hardcoded but:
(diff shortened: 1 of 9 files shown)


plugin/src/openshift/utils/KialiIntegration.ts
```diff
@@ -26,6 +27,14 @@ export const getPluginConfig = async function (): Promise<PluginConfig> {
   });
 };
 
+// Set the router basename where OSSMC page is loaded
+export const setRouterBasename = (pathname: string): void => {
+  const ossmConsoleIndex = pathname.indexOf('/ossmconsole');
```
josunect, line 32: I think that this plugin base name could be better defined as a constant, so it could be identified easily in case of change or used in other places.

josunect (commented):
The refresh of the Service Mesh page works as expected: 

image

But there are other routes that are incorrect, like <Create Traffic Policies> from the Overview: 

image

image

Or disable auto injection: 

image

Probably they are not related with this change but I think they should be fixed (Probably in a separate PR), or disabled

aljesusg:
We should backport this to 1.86 too

ferhoyos:
Yes, I will create the backport to 1.86 when this PR is reviewed and merged

plugin/src/openshift/utils/KialiIntegration.ts
ferhoyos: do you mean define `/ossmconsole` as a constant?

ferhoyos:
Thanks for the review @josunect. It is not related with this PR, but I have fixed the `Overview` link issues you mention.

plugin/src/openshift/utils/KialiIntegration.ts
josunect: Yes, right

josunect (approved):
Now these link are also fixed:

image

LGTM

Issue #325, opened 2024-06-28
Service details tab navigation gets stuck after a refresh


ferhoyos:
### Describe the bug

If the user goes to a Service details page, clicks on the `Service Mesh` tab, and refreshes the page, the Service Mesh details page gets stuck on the service mesh content info even though the active tab is the default `Details` tab. In this situation, the tab navigation is broken, and the user cannot see the rest of the Service tabs

### Expected Behavior

After a refresh, the page should display the same content as before the refresh, and the tab navigation should work as expected.

### What are the steps to reproduce this bug?

1. Go to any service details page
2. Click on the `Service Mesh` tab
3. Refresh the page
4. Verify that the service mesh tab content is shown even though the active tab is the `Details` one.

![image](https://github.com/kiali/openshift-servicemesh-plugin/assets/122779323/cba54fc6-8123-42a0-94cf-95ca3bdc61ed)

5. Verify that nothing is displayed when the user clicks on other tabs of the service details page.

![image](https://github.com/kiali/openshift-servicemesh-plugin/assets/122779323/bd189c7f-bd58-4665-a7d3-41c04b911b49)

### Environment

- **OSSMC/Kiali version:** Latest
- **OpenShift version:** 4.15
- **Istio version:** 1.22

ferhoyos:
The issue only happens with OpenShift 4.15+ and also affects v1.73.

PR #333, opened 2024-07-04
[v1.86] Service mesh tab is loaded correctly after refresh (#327)

ferhoyos:
Backport of PR #327 to version 1.86

Issue: #325


josunect (approved):
LGTM
image

I think there should be an issue the indicate the need to translate the Chinese missing strings

PR #338, opened 2024-07-29
Add PR option to copy kiali repo script

ferhoyos:
### Describe the change

To test a Kiali PR that affects OSSMC, the hack script to copy the Kiali repo into OSSMC needs an option to get the code from a PR instead of a branch or tag reference (e.g., https://github.com/kiali/kiali/pull/7568).

### Steps to test the PR

1- Execute the hack script with `-pr` option and any valid PR number from Kiali repo. For example:

```
${OSSMC_DIR}/hack/copy-frontend-src-to-ossmc.sh -pr 7568
```

2 - Verify that the changes for that specific PR have been copied into the OSSMC repo

### Automation testing

N/A

### DIFF
--- a/hack/copy-frontend-src-to-ossmc.sh
+++ b/hack/copy-frontend-src-to-ossmc.sh
@@ -34,6 +34,7 @@ SOURCE_REPO_URL="${DEFAULT_SOURCE_REPO_URL}"
 # The git ref (branch or tag name) to checkout when cloning the source repo
 DEFAULT_SOURCE_REF="master"
 SOURCE_REF="${DEFAULT_SOURCE_REF}"
+GIT_REF="${SOURCE_REF}"
 
 # This is to be the top-level directory of the local OSSM git repo.
 # This is where DEST_DIR should be located.
@@ -44,6 +45,7 @@ while [[ $# -gt 0 ]]; do
   key="$1"
   case $key in
     -dr|--dest-repo)     DEST_REPO="$2"         ;shift;shift ;;
+    -pr|--pull-request)  PULL_REQUEST="$2"      ;shift;shift ;;
     -sr|--source-ref)    SOURCE_REF="$2"        ;shift;shift ;;
     -su|--source-url)    SOURCE_REPO_URL="$2"   ;shift;shift ;;
     -h|--help)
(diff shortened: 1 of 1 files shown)


hack/copy-frontend-src-to-ossmc.sh
```diff
@@ -44,6 +45,7 @@ while [[ $# -gt 0 ]]; do
   key="$1"
   case $key in
     -dr|--dest-repo)     DEST_REPO="$2"         ;shift;shift ;;
+    -pr|--pull-request)  PULL_REQUEST="$2"      ;shift;shift ;;
```
jmazzitelli, line 48: add a brief --help message down between lines 61 and 63 to let people know what this option is.

hack/copy-frontend-src-to-ossmc.sh
```diff
@@ -34,6 +34,7 @@ SOURCE_REPO_URL="${DEFAULT_SOURCE_REPO_URL}"
 # The git ref (branch or tag name) to checkout when cloning the source repo
 DEFAULT_SOURCE_REF="master"
 SOURCE_REF="${DEFAULT_SOURCE_REF}"
+GIT_REF="${SOURCE_REF}"
```
jmazzitelli, line 37: This should move down after the options are processed (i.e. below line 85). Otherwise, GIT_REF will always be the ${DEFAULT_SOURCE_REF} when `-pr` is not used. You want GIT_REF to pick up SOURCE_REF that might be changed when someone passes in `-sr`.

jmazzitelli (changes requested):
I have some requested changed in other comments.

Also, this is breaking previous functionality. I actually had to use this script just last week. I was preparing a demo of the OSSMC plugin using code that was not in kiali/kiali repo - it was in jay's branch.

So, I ran this (which was successful last week):

```
./hack/copy-frontend-src-to-ossmc.sh -su https://github.com/jshaughn/kiali.git -sr kiali7485
```

But now that fails with this:

```
Cloning source repo [https://github.com/jshaughn/kiali.git] (branch/tag [kiali7485]) into [/tmp/KIALI_OSSM.tmp/kiali-source].
./hack/copy-frontend-src-to-ossmc.sh: line 118: PULL_REQUEST: unbound variable
```

I think to fix this, just define PULL_REQUEST below line 37 so it is:

```
PULL_REQUEST=""
```

ferhoyos:
Thanks for the review @jmazzitelli.  I think I have fixed all the issues you mention and added the brief help.

Looking at your example I have noticed that `-pr` option is a mix of `-sr` and `-su` since a pull request comes from a branch in the developer code repository. Maybe it is good to have both options so the developer can choose.

jmazzitelli:
> Maybe it is good to have both options so the developer can choose.

Yeah, that's fine. Both should work. I think if you make those changes I suggest, it will.

jmazzitelli (approved):
This works. I tried with the new `-pr` and I tried also with `-su`/`-sr` and all worked, too.

PR #342, opened 2024-08-01
Update dev dependencies

ferhoyos:
### Describe the change

This PR updates some frontend dev dependencies that do not affect Kiali runtime.

List of updates fixed:
- [OSSM-6575](https://issues.redhat.com/browse/OSSM-6575)
- [OSSM-6647](https://issues.redhat.com/browse/OSSM-6647)

### Steps to test the PR

1. Verify that CI tests pass

### Automation testing

N/A

PR #347, opened 2024-08-08
[v1.86] Update dev dependencies and micromatch

ferhoyos:
Backport of PRs https://github.com/kiali/openshift-servicemesh-plugin/pull/342 and https://github.com/kiali/openshift-servicemesh-plugin/pull/344 to version 1.86

### DIFF
--- a/plugin/package.json
+++ b/plugin/package.json
@@ -113,7 +113,7 @@
     "file-loader": "^6.2.0",
     "fork-ts-checker-webpack-plugin": "^8.0.0",
     "husky": "^8.0.3",
-    "i18next-parser": "^8.12.0",
+    "i18next-parser": "^9.0.1",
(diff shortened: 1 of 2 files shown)


plugin/package.json
aljesusg, line 1: Can you add in resolutions the micromatch version?

"micromatch": "^4.0.7",

aljesusg (changes requested):

plugin/package.json
ferhoyos: There is no problem because the `yarn.lock` file is the only source of truth and it will always download `^4.0.7`, even for a fresh yarn install.

Resolutions are needed when you have a conflict with major or fixed versions of a library (e.g., you have versions 2.X and 3.X of the same library but you want to use only one of them). For patch versions like micromatch, a dependency update in the `yarn.lock` file is enough.

Issue #356, opened 2024-08-13
Reuse Kiali cypress tests for OSSMC


ferhoyos:
### What do you want to improve?

Reuse Kiali tests for OSSMC

### What is the current behavior?

Currently, we assume that cypress tests executed for Kiali are valid for OSSMC as well, but some recent issues that only affect OSSMC makes this assumption invalid. 

### What is the new behavior?

The new approach mirrors how we handle the Kiali code. All Kiali Cypress tests are copied into the test folder every sprint, and by configuration, we select which tests are applicable to OSSMC.

Some tasks have to be addressed for this epic:

- [X] https://github.com/kiali/openshift-servicemesh-plugin/pull/373
- [X] https://github.com/kiali/openshift-servicemesh-plugin/pull/357
- [X] https://github.com/kiali/kiali/pull/7656
- [X] https://github.com/kiali/openshift-servicemesh-plugin/pull/376
- [x] [Include OSSMC tests in nightly upstream job](https://gitlab.cee.redhat.com/istio/servicemesh-qe/jenkins-csb-declaration/-/merge_requests/515)

ferhoyos:
The epic to reuse the Kiali tests in OSSMC is completed :tada:

Issue #330, opened 2024-07-02
OSSMC cannot update namespace or create Istio objects


ferhoyos:
### Describe the bug

An error appears when the user tries to update a namespace (e.g., enable/disable injection) or create any Istio object (e.g., create traffic policies).

![image](https://github.com/kiali/openshift-servicemesh-plugin/assets/122779323/766bfb61-6bfe-49b3-b783-925ad8d45010)

### Expected Behavior

The user should update any namespace or create any Istio object from OSSMC plugin

### What are the steps to reproduce this bug?

1. Go to the overview page
2. Click on the context menu of any namespace
3. Click on enable/disable injection option
4. Verify that an error appears and the namespace is not updated.

### Environment

- **OSSMC/Kiali version:** Latest
- **OpenShift version:** 4.15
- **Istio version:** 1.22

ferhoyos:
@ScriptingShrimp found that the issue still happens, but for a different reason (the original issue caused a `500 Internal Server Error`, but now it results in a `403 Forbidden`).

PR #357, opened 2024-08-19
Run Kiali tests in OSSMC

ferhoyos:
### Describe the change

When we reused Kiali's source code as the foundation for the OSSMC plugin, we assumed that Kiali's Cypress tests would also be valid for OSSMC. In that case, we only needed to create Cypress tests for specific OSSMC pages, such as the Istio config list page or specific navigation scenarios. However, recent issues (see[ issue #330](https://www.github.com/kiali/openshift-servicemesh-plugin/issues/330)) have shown that the same test scenarios can behave differently in Kiali and OSSMC.

To address this, we have modified the OSSMC Cypress test infrastructure to run the same Kiali test scenarios in OSSMC. We have introduced the `@ossmc` Cypress tag to run only those Kiali tests that are applicable to OSSMC (excluding multi-cluster, wizards, etc.).

Actions performed to enable running Kiali tests:
- Defined the same environment variables used in Kiali Cypress tests (`USERNAME`, `PASSWD`, `AUTH_PROVIDER`, etc.)
- Created the `@ossmc` Cypress tag to include a Cypress test in the OSSMC test suite.
- Created the `@skip-ossmc` Cypress tag to exclude specific Cypress tests from features included in OSSMC.
- Organized Kiali and OSSMC Cypress tests into their respective test folders (similar structure to the code, with`kiali` and `openshift` folders)
- Overrode the `visit` and `request` Cypress commands since there are slight differences in page URLs and APIs between Kiali and OSSMC.

### Steps to test the PR

1. Execute `yarn cypress:run` and verify that OSSMC cypress test pass

### Issue reference

https://github.com/kiali/openshift-servicemesh-plugin/issues/356


ScriptingShrimp:
testing via upstream-istio-pipeline/2165

plugin/cypress/README.md
```diff
@@ -12,9 +12,9 @@ Before you start using Cypress suite, you might need export some environment var
 
 ```bash
 export CYPRESS_BASE_URL=<value>               # defaults to http://localhost:9000
-export CYPRESS_OC_CLUSTER_USER=<value>        # defaults to jenkins, opt. kubeadmin
-export CYPRESS_OC_CLUSTER_PASS=<value>        # no defaults
-export CYPRESS_OC_IDP=<value>                 # defaults to my_htpasswd_provider
+export CYPRESS_USERNAME=<value>               # defaults to jenkins, opt. kubeadmin
```
ScriptingShrimp, line 15: Originally I chose different variable names to be double sure that settings from kiali/kiali will not interfere with OSSMC suite

plugin/cypress.config.ts
```diff
@@ -18,13 +18,14 @@ export default defineConfig({
   videosFolder: 'cypress/videos',
 
   env: {
-    OC_CLUSTER_USER: 'jenkins', // default value for jenkins
-    OC_IDP: 'my_htpasswd_provider', // default value for jenkins, can vary based on cluster setup
     'cypress-react-selector': {
       root: '#root'
     },
+    OSSMC: true,
```
ScriptingShrimp, line 24: only nitpick, but this is why I would prefer to have two different configs, so there is no risk for them to be overwritten when copying the code

ScriptingShrimp (approved):
This is great!

plugin/cypress.config.ts
ferhoyos: Actually there are two configs, the `cypress.config` from Kiali repo and `cypress.config` from OSSMC repo. What I don't see is to have two configs just for OSSMC.

PR #364, opened 2024-09-05
Remove install cypress test for OSSMC

ferhoyos:
### Describe the change

OSSMC contains 2 cypress tests (`install_ossmc.feature` and `uninstall_ossmc.feature`) that were created before any other cypress test, just to check that at least OSSMC was installed and uninstalled correctly in the OpenShift console. Currently they are not needed because if the OSSMC installation goes wrong, the other OSSMC cypress tests will fail.

### Steps to test the PR

Execute `yarn cypress:run` and verify that the cypress tests pass

### Automation testing

Remove install and uninstall cypress test scenarios


ScriptingShrimp (approved):
I was able to deploy OSSMC using `make deploy-plugin enable-plugin` as described in README. Regression passing OK, LGTM!

PR #376, opened 2024-09-13
[v1.73] Replace default cypress env variables

ferhoyos:
### Describe the change

Partial backport of #357 where we replace the default user, password and auth provider cypress environment variables.

### Steps to test the PR

Verify that the Cypress tests run successfully

### Issue reference

https://github.com/kiali/openshift-servicemesh-plugin/issues/356


ScriptingShrimp (approved):
LGTM!

PR #378, opened 2024-09-17
Update serialize-javascript dependency

ferhoyos:
### Describe the change

Update `serialize-javascript` dependency to version 6.0.2.

Instead of adding this version to `resolutions` section, I have updated the parent dependency that import `serialize-javascript` as it was a bit outdated. This will improve the application's security."

Library updated:
- `@badeball/cypress-cucumber-preprocessor` to version `20.1.2`.  

Breaking changes in the major version do not affect us:
- https://github.com/badeball/cypress-cucumber-preprocessor/releases?page=2

**Bonus:** 
- Align `seed-random` version in `package.json` to download only one version in the application.

### Steps to test the PR

Verify that CI tests pass.

### Issue reference

https://issues.redhat.com/browse/OSSM-8145

### DIFF
--- a/plugin/package.json
+++ b/plugin/package.json
@@ -77,7 +77,7 @@
   },
   "devDependencies": {
     "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
-    "@badeball/cypress-cucumber-preprocessor": "^19.2.0",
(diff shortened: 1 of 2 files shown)


aljesusg (approved):
LGFM

PR #385, opened 2024-10-15
[v1.73] Update dompurify dependency

ferhoyos:
### Describe the change

Updates the `dompurify` dependency in OSSMC version 1.73. This dependency is not used in the current `main` branch, so there is no need to update it there.

### Steps to test the PR

Verify that OSSMC works as expected

### Automation testing

N/A

### Issue reference

https://issues.redhat.com/browse/OSSM-8246


aljesusg (approved):
LGFM

PR #390, opened 2024-10-23
Ability to do major releases

leandroberetta:
Ability to do major releases.

Scheduled releases are disabled for now, I want to test minor releases this week before activating them again.


ferhoyos (approved):
LGTM

PR #427, opened 2025-03-19
Patch release fix

leandroberetta:
Fix an issue that sets wrong versions for patch releases and also add some steps to create a PR in case of patch releases, to update the version branch (ie v2.7).

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -244,6 +244,8 @@ jobs:
       run: |
         docker login -u ${{ secrets.QUAY_USER }} -p ${{ secrets.QUAY_PASSWORD }} quay.io
 
+        make build-push-plugin-multi-arch
+
     - name: Create tag
       run: |
         git add Makefile plugin/package.json plugin/plugin-metadata.ts
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -45,7 +45,7 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v4
       with:
-        ref: ${{ github.ref_name }}
+        ref: ${{ github.event.inputs.release_branch || github.ref_name }}
 
     - name: Prepare scripts
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -260,9 +259,40 @@ jobs:
...
+      run: |
+        if git ls-remote --exit-code --heads origin "$BRANCH_VERSION"; then
+          echo "EXISTS=true" >> $GITHUB_ENV
+        else
+          echo "EXISTS=false" >> $GITHUB_ENV
+        fi
+
+    - name: Create version branch
+      if: ${{ env.EXISTS == 'false' && needs.initialize.outputs.release_type == 'patch' }}
       run: git push origin $(git rev-parse HEAD):refs/heads/$BRANCH_VERSION
```
ferhoyos, line 273: Do we have a use case when the release type is `patch` and the the version branch is not created? If the answer is no, I prefer to throw an error because this means something wrong is happening.

leandroberetta: I added this check for this PR only, thought it was an improvement, but really, when branch doesn't exist, it means is the first time creating the version branch, meaning, a minor release. 

So, both cases are going to happen, when false, it's a minor, and the branch needs to be created.

When true, it already exists, it's a patch release. I'm going to remove the and condition that checks the release type, because the false is already enought I think.

leandroberetta: Well... I changed the approach a little bit, you'll notice that:

1. if it is a minor/major release, branch won't exist, needs to be created
2. if it is a patch release, create a PR directly

.github/workflows/release.yaml
```diff
@@ -260,21 +258,23 @@ jobs:
...
+        BUILD_TAG: kiali-release-${{ github.run_number }}-${{ needs.initialize.outputs.branch_version }}
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      if: ${{ needs.initialize.outputs.release_type == 'minor' || needs.initialize.outputs.release_type == 'major' }}
       run: |
-        hack/update-version-string.sh "${NEXT_VERSION}-SNAPSHOT"
-
-        git add Makefile plugin/package.json plugin/plugin-metadata.ts
-
-        git commit -m "Prepare for next version"
+        if [[ $RELEASE_TYPE == "major" || $RELEASE_TYPE == "minor" ]]; then
```
ferhoyos, line 270: I have a question here for patch releases: I see that `hack/update-version-string.sh "$RELEASE_VERSION"` and `git add Makefile plugin/package.json plugin/plugin-metadata.ts` are already executed in previous steps. Do we need to execute `git commit` for patch releases or the `gh release` command does it for us?

ferhoyos (commented):
@leandroberetta the code looks good. I have two questions to understand better how the release github workflow works before approval.

.github/workflows/release.yaml
leandroberetta: Hi @ferhoyos, in the case of:

- Patch releases: the already executed script that sets versions up is enough to create the PR. Meaning, new versions ofr the vx.y branch will have the version which is being released. 
- Minor/Major releases, the script needs to be run again with the next version parameter, to prepare the main branch for the upcoming version, for that reason there is the if else there.

ferhoyos: Yes, but for the patch releases I don't see a git push or git commit command that push the release version into the branch (only git add). I'm just asking if the gh release command performs such action

leandroberetta: All the work is done in previous steps when creating the tag, when creating the tag:

- The versions are updated
- Files are added
- Tag is created

Then, just for the case of major/minor, versions are updated again (and added, commited)

The PR is created for all cases, patch releases don't require the next version, usually the branch version is ok to have the released version.

ferhoyos (approved):
Ok, thanks for the clarification @leandroberetta . It looks fine to me, approved!

PR #429, opened 2025-03-26
Remove cytoscape graph impl

jshaughn:
Part-of https://github.com/kiali/kiali/issues/8053


plugin/manifest.yaml
```diff
@@ -118,7 +118,6 @@ metadata:
 data:
   plugin-config.json: |
     {
```
ferhoyos, line 120: I think we should put here the default value from `plugin/plugin-config.json`: 
```
plugin-config.json: |
    {
        "observability": {
            "instance": "sample",
            "namespace": "tempo",
            "tenant": "default"
          }
     }
```
@josunect could you please confirm?

plugin/src/openshift/components/KialiController.tsx
```diff
@@ -79,11 +79,7 @@ type KialiControllerProps = KialiControllerReduxProps & {
   children: React.ReactNode;
 };
 
-const defaultPluginConfig: PluginConfig = {
-  graph: {
-    impl: 'pf'
-  }
-};
+const defaultPluginConfig: PluginConfig = {};
```
ferhoyos, line 82: @josunect should we add the observability default config here as well?
```suggestion
const defaultPluginConfig: PluginConfig = {
  "observability": {
    "instance": "sample",
    "namespace": "tempo",
    "tenant": "default"
  }
}
```

ferhoyos (commented):
The code looks good to me, I will give the approval after checking the value of the default plugin config.

plugin/manifest.yaml
jshaughn: I think you are right (in both places).  Since Josune is busy at KubeCon I think we could probably make the addition and remove it if we are wrong.  Let me know.

PR #434, opened 2025-04-04
Change overview request API in the cypress test

ferhoyos:
### Describe the change

The cypress test `user sees istio-system overview card` does not work anymore because the istio status API is not longer called when OSSMC loads the overview page. This PR calls to istio validations API instead to check that overview page is correctly loaded.

### Steps to test the PR

Verify that CI tests pass


hhovsepy (approved):
thanks @ferhoyos

PR #443, opened 2025-05-19
Add Shared Mesh Config Cypress hook to OSSMC

ferhoyos:
### Describe the change

Adapt Kiali PR https://github.com/kiali/kiali/pull/8330 to OSSMC since the cypress hooks are different

### Steps to test the PR

Verify that CI tests pass

### Issue reference

https://github.com/kiali/kiali/issues/8248


nrfox (approved):

nrfox:
@ferhoyos why are the hooks duplicated?

ferhoyos:
Because the demo installation is a bit different in OSSMC and Kiali. Hack scripts in OSSMC are downloaded in runtime from Kiali repo to avoid another copy to OSSMC repo (`hack/download-hack-scripts.sh`). These scripts change less frequent than cypress tests or code. If we copied the hack scripts to OSSMC, I think the hook file could be the same.

Issue #447, opened 2025-06-03
Guided tour on OCP 4.19+


mkralik3:
### Describe the bug

On OCP 4.19+, there is a new Guided tour wizard right after the login. 
There is no global value for disabling it. The only way how to do that is **login first** and then:

- in the UI, click on Skip button
- via `oc`, find user CM in `openshift-console-user-settings` namespace, update `.data.console.guidedTour: '{"admin":{"completed":true}}'` and then refresh the UI

![Image](https://github.com/user-attachments/assets/063a9887-7c31-4b79-9f1a-30f1e424f671)

mkralik3:
1.73 -> not supported on OCP 4.19+
2.4 -> https://github.com/kiali/openshift-servicemesh-plugin/pull/448
mains -> https://github.com/kiali/openshift-servicemesh-plugin/pull/449

PR #448, opened 2025-06-03
[2.4] Fix guided tour after first login on OCP 4.19+

mkralik3:
### Describe the change

Fix for https://github.com/kiali/openshift-servicemesh-plugin/issues/447


ScriptingShrimp:
Hey @mkralik3 , thanks for proposal. If I understand correctly, this adds 5sec timer into each test case that fires the `guidedTour.close()` function. Is that correct?

mkralik3:
@ScriptingShrimp  yes, unfortunetelly, `cy.waitForReact` doesn't work there. It is timeouted on OCP home page. I have tried also with big value `cy.waitForReact(90000)`

mkralik3:
Verified that login works on OCP 4.19 and it sloses the guided window (i was not able to try the whole test suite due to the console slowness issue which was still not resolved)

ferhoyos (approved):
LGTM

PR #485, opened 2025-09-17
Fix pod handling for workloads without version

josunect:
### Describe the change

ossmc does a redirection to the workload based on the url pod, but it is not taking into account the pods without the version: 

- istiod-866fd6ccd7-7v8p5
- details-v1-77b775f46-c7vjb

In the case of Istio, tries to look for the workload istiod-866fd6ccd7 which is not found.

### Steps to test the PR

- Run kiali
- Run ossmc, can be run locally: https://github.com/kiali/openshift-servicemesh-plugin/?tab=readme-ov-file#run-the-plugin-and-openshift-console-locally
- Go to a Istio pod, it should work. It should also work for other pod including the version, like the bookinfo

### Automation testing

Url used by ambient tests (In progress)

### Issue reference

Fixes https://github.com/kiali/kiali/issues/8728

### DIFF
--- a/plugin/src/openshift/pages/MeshTab/WorkloadMesh.tsx
+++ b/plugin/src/openshift/pages/MeshTab/WorkloadMesh.tsx
@@ -25,28 +25,21 @@ const WorkloadMeshTab: React.FC<void> = () => {
   if (ns && name && plural) {
     let workload = name;
     if (plural === 'pods') {
-      let count = 0;
-      let index = 0;
+      const parts = workload.split('-');
 
-      // Get the parent workload (app-version) from pod identifier
-      for (let i = 0; i < workload.length; i++) {
-        if (workload[i] === '-') {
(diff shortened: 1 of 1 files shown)


ferhoyos (commented):
Nice catch @josunect. Your solution works, but I think that we can simplify the whole logic removing the last two hyphen-separated segments instead of trying to figure out if the pod name includes the workload version or not.

```
if (plural === 'pods') {
    // Remove the last two hyphen-separated segments from pod name to get workload name
    // e.g., istiod-866fd6ccd7-7v8p5 -> istiod
    // e.g., details-v1-77b775f46-c7vjb -> details-v1
    const parts = workload.split('-');
    if (parts.length > 2) {
      workload = parts.slice(0, -2).join('-');
    }
  }
```

ferhoyos (approved):

PR #491, opened 2025-09-23
Validate workload name in the backend

josunect:
### Describe the change

Validate workload name in the backend

### Steps to test the PR
Apply different pods and go to the OpenShift pod service mesh tab to validate they are found: 
`hack/istio/ambient/install-waypoints.sh `

```
apiVersion: v1
kind: Pod
metadata:
  name: my-pod
  labels:
    app: test-pod
spec:
  containers:
    - name: test-pod
      image: curlimages/curl:latest
      command: [ "sleep", "3600" ]  
  restartPolicy: Never
```

Try different pod names, like mypod, my-pod-v1

### Automation testing

If applicable, explain the case scencarios covered by **e2e** tests created for this PR.

### Issue reference
Fixes https://github.com/kiali/kiali/issues/8749

### DIFF
--- a/plugin/src/openshift/pages/MeshTab/WorkloadMesh.tsx
+++ b/plugin/src/openshift/pages/MeshTab/WorkloadMesh.tsx
@@ -82,7 +82,20 @@ const WorkloadMeshTab: React.FC<void> = () => {
       // Only validate if the parsed name is different from the original
       if (parsedWorkload !== name) {
         const isValid = await validateWorkload(ns, parsedWorkload);
-        setFinalWorkload(isValid ? parsedWorkload : name);
+        if (!isValid) {
+          // If validation fails, try with just removing the last part
+          // e.g. kiali-traffic-generator-n2d9n -> kiali-traffic-generator
+          const parts = name.split('-');
+          if (parts.length >= 3) {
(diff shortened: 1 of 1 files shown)


ferhoyos (changes requested):
Great PR @josunect! and sorry for the delay in the review. It is great that workload is validated before loading. Just one suggestion: make the parseWorkload method recursive until `parsedWorkload === name`. I mean, for `parts.length >= 3`, you first remove two last parts. If it is not valid, I think you could try removing only the last part and call validation again (e.g. for pod `kiali-traffic-generator-6q9rx`). If it is not valid again, the next time `parsedWorkload === name` and this the final name.

josunect:
> Great PR @josunect! and sorry for the delay in the review. It is great that workload is validated before loading. Just one suggestion: make the parseWorkload method recursive until `parsedWorkload === name`. I mean, for `parts.length >= 3`, you first remove two last parts. If it is not valid, I think you could try removing only the last part and call validation again (e.g. for pod `kiali-traffic-generator-6q9rx`). If it is not valid again, the next time `parsedWorkload === name` and this the final name.

@ferhoyos I've made a commit with those changes

ferhoyos (approved):
Thanks for the quick fix @josunect. LGTM!

PR #647, opened 2026-04-24
Upgrade follow-redirects to 1.16.0

ferhoyos:
## Summary

- Upgrades `follow-redirects` from 1.15.11 to 1.16.0 to fix CVE-2026-40895
- `follow-redirects` prior to 1.16.0 forwards custom authentication headers (X-API-Key, X-Auth-Token, etc.) to cross-domain redirect targets, resulting in information disclosure
- Transitive dependency — only `yarn.lock` is modified

## Test plan

- [x] `yarn install` completes successfully


jshaughn (approved):
LG, CI Passed.

PR #672, opened 2026-05-07
Adopt Rspack-aligned tooling (SWC + Rstest) for Kiali rsbuild migration

ferhoyos:
## Summary

Kiali is moving to rsbuild/rspack for its build and test tooling. The OpenShift Console SDK does not yet support rspack, so OSSMC must stay on webpack for now. However, we can adopt rspack-compatible tooling within webpack to minimize the future migration effort.

This PR replaces Babel with SWC for builds and Jest with Rstest for tests. Both are the same Rust-based tools that rspack/rsbuild uses internally, so:

- OSSMC builds and tests are now aligned with Kiali's Rstack toolchain
- Future rspack migration (once the SDK supports it) becomes minimal
- Build times improve ~3-5x compared to Babel, tests are faster than Jest

### Build changes (SWC)

- Replace `babel-loader` + `babel-preset-react-app` with `swc-loader` + `@swc/core`
- Remove `@babel/plugin-proposal-private-property-in-object` (no longer needed without Babel)
- SWC options match rspack's `builtin:swc-loader` API for easy future swap
- Keep `fork-ts-checker-webpack-plugin` for type checking (SWC does not type-check)

### Test changes (Rstest)

- Replace Jest + `@swc/jest` + `jest-environment-jsdom` with `@rstest/core` + `jsdom`
- Create `rstest.config.ts` with jsdom environment, globals, tsconfig path resolution
- Migrate mock files and tests from `jest.mock`/`jest.fn` to `rstest.mock`/`rstest.fn`
- Replace mutable `require()` mock pattern with `rstest.hoisted()` + getters for ESM compatibility
- Remove `@types/jest`, `jest-canvas-mock`, `jest-localstorage-mock`
- All 58 tests pass, aligned with Kiali PR kiali/kiali#9653

## Test plan

- [x] `yarn build:dev` compiles successfully
- [x] `yarn test` -- all 3 suites, 58 tests pass with Rstest
- [ ] `yarn start` dev server works with HMR
- [ ] Production build (`yarn build`) produces working plugin
- [ ] Plugin loads correctly in OpenShift Console


jshaughn (approved):
@ferhoyos Looks good with the latest update of kiali `src`. Builds for me and plugin runs normally from what I can see (using CRC 2.61 which is k8s 1.34 OCP 4.21).

One non-blocking comment. In a follow-up you probably want to ensure AGENTS.MD doesn't have any dated references (I see at least one reference to Jest, for example).

PR #680, opened 2026-05-26
Fix immutable.js v3/v4 type conflict causing build errors

ferhoyos:
## Summary
- Add `"immutable": "^4.3.8"` to `resolutions` in `package.json` to force all consumers to use immutable v4, fixing 7 TypeScript build errors in vendored ChatBot code.
- Fix the `aiChat` reducer key name in `Reducer.ts` to match the `KialiAppState` interface.

### Root cause
`@openshift-console/dynamic-plugin-sdk` depends on `immutable@3.x`, which Yarn hoists to the top level. The v3 type definitions declare `map()`/`filter()` callback parameters as optional (`value?: V, key?: K`), making them `T | undefined` under `strictNullChecks`. The Kiali ChatBot code was written against v4 types where these are non-optional. The SDK's runtime usage is trivial (`Map()` constructors only) and fully compatible with v4.

## Test plan
- [ ] `yarn install` resolves only `immutable@4.3.8` (verify with `yarn why immutable`)
- [ ] `yarn build` completes without the 7 previous TypeScript errors
- [ ] Smoke test the plugin in OpenShift Console to verify SDK functionality is unaffected


aljesusg (approved):
LGFM

PR #687, opened 2026-05-28
update workflows

jmazzitelli:
This PR hardens the GitHub Actions workflows against several known vulnerability classes.

**Supply chain risk (unpinned actions):** All third-party actions were using floating version tags (e.g. `@v3`, `@v4`, `@v5`). These are mutable references — a compromised action maintainer can silently retag them to malicious code that executes in CI with access to repository secrets. All action references have been pinned to their full immutable commit SHA, with the version tag retained as a comment.

**Credential exposure via `docker login -p`:** The nightly and release workflows were passing Quay credentials directly as a `-p` CLI argument to `docker login`, exposing the password in the process list (`ps aux`). Both have been replaced with `docker/login-action`, which passes credentials securely. The CI and test-images workflows were already using `docker/login-action` correctly.

**Missing `permissions:` blocks:** None of the workflows declared explicit token permissions, meaning they inherited the repository default (typically broad write access). All workflows now declare `permissions: read-all` at the top level, with `contents: write` and/or `pull-requests: write` granted only to the specific jobs that actually need them (release, tag creation, version bumping, branch pushes).

**Unquoted shell variable expansion:** The release workflow had an unquoted `[ -z ${{ }} ]` expression that could behave unpredictably with empty inputs. This has been quoted.

Generated-by: Claude

### DIFF
--- a/.github/workflows/bump-release-version.yml
+++ b/.github/workflows/bump-release-version.yml
@@ -53,7 +53,7 @@ jobs:
         branch: ${{fromJson(needs.initialize.outputs.branches)}}
     steps:
     - name: Checkout branch
-      uses: actions/checkout@v5
+      uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd  # v5
       with:
         ref: ${{matrix.branch}}
         # We need to fetch the full history to determine the latest tag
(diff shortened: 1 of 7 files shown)


jshaughn (approved):
I think this is all fine. We'll see.

PR #701, opened 2026-06-09
[v2.27] Fix waypoint table headers missing due to uninitialized i18next singleton

### LINKED ISSUE #699: Waypoint table column headers are not displayed in OSSMC
## Description

In OSSMC, the Waypoint tab (under the Service Mesh tab for workload details) renders the enrolled services/workloads table without visible column headers. The table data rows appear correctly, but the `<Thead>` column names (Name, Namespace, Labeled by) are invisible.

This works correctly in the standalone Kiali application — column headers are fully visible.

## Steps to Reproduce

1. Deploy a waypoint proxy workload in an ambient mesh namespace (e.g., `cgw` in namespace `waypoint-forall`)
2. In the OpenShift Console, navigate to the workload's detail page
3. Click the **Service Mesh** tab, then the **Waypoint** sub-tab
4. Observe the **Services** table under "List of enrolled services"

## Expected Behavior

The table should display column headers: **Name**, **Namespace**, **Labeled by** — as it does in standalone Kiali.

## Actual Behavior

The column header row is present in the DOM but the text is not visible. Only the data rows are displayed.

Image

## Environment

- OSSMC plugin (OpenShift Console dynamic plugin)
- Ambient mesh with waypoint proxies deployed

ferhoyos:
## Summary

Backport of #700 to v2.27.

- Initialize the default `i18next` singleton in OSSMC so that vendored Kiali code importing `t` directly from `'i18next'` returns key fallbacks instead of `undefined`
- Remove the redundant `fallbackI18n` instance, reusing the now-initialized default singleton
- Switch `MenuTitles.ts` to use `utils/I18nUtils` for consistency

## Root Cause

Vendored components like `WaypointWorkloadsTable` and `WaypointLabel` import `t` directly from `'i18next'` instead of `'utils/I18nUtils'`. In OSSMC the default `i18next` singleton was never initialized -- only a separate `createInstance()` copy was. Calling `t()` on an uninitialized instance returns `undefined`, causing column headers and translated strings to disappear.

Fixes #699

### DIFF
--- a/plugin/src/openshift/i18n.ts
+++ b/plugin/src/openshift/i18n.ts
@@ -1,15 +1,18 @@
 import i18next, { i18n as I18nInstance } from 'i18next';
 import { getI18n } from 'react-i18next';
 
-const fallbackI18n = i18next.createInstance();
-void fallbackI18n.init({ fallbackLng: 'en' });
+// Initialize the default i18next singleton so that vendored Kiali code
+// importing `t` directly from 'i18next' gets key fallbacks instead of undefined.
+if (!i18next.isInitialized) {
+  void i18next.init({ fallbackLng: 'en' });
+}
 
-const resolveI18n = (): I18nInstance => getI18n() ?? fallbackI18n;
+const resolveI18n = (): I18nInstance => getI18n() ?? i18next;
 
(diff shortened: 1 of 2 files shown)


josunect (approved):
LG

PR #776, opened 2026-07-15
fix: correct pretty-quick --pattern syntax and add .prettierignore

ferhoyos:
## Summary

- Fix the `--pattern` bracket syntax in the pre-commit script that has been silently causing `pretty-quick` to format zero files
- Add `.prettierignore` to exclude vendored `src/kiali/` from formatting
- Fix missing trailing newline in `.husky/pre-commit`

## Problem

The pre-commit hook used `--pattern ['**/*.{ts,tsx,scss,json}', '!src/kiali']` which is not valid CLI syntax for pretty-quick. After shell parsing, `mri` receives `[**/*.{ts,tsx,scss,json},` as the pattern value, and both `multimatch` (v3) and `picomatch` (v4) fail to match any file paths against it. This means `pretty-quick` has been silently skipping formatting on every commit.

Verified with picomatch v4:
```
Pattern "[**/*.{ts,tsx,scss,json}," matches:
  src/openshift/components/Foo.tsx: false
  src/kiali/store/Store.ts: false
  config.json: false
```

## Fix

1. **Aligned pattern syntax with upstream kiali/kiali**: `--pattern "**/*.{ts,tsx,scss,json}"`
2. **Added `plugin/.prettierignore`** with `src/kiali/` to exclude vendored code (the previous `!src/kiali` exclusion via `--pattern` never worked and picomatch doesn't support negation in pattern arrays as "exclude" logic anyway)
3. **Fixed missing trailing newline** in `.husky/pre-commit` (violates `.editorconfig`)

## Test plan

- [ ] Run `yarn install && yarn pre-commit` — verify pretty-quick now actually formats staged files
- [ ] Stage a file in `src/kiali/` — verify it is NOT formatted by pretty-quick (excluded via `.prettierignore`)
- [ ] Run `yarn prettier --check src/openshift/` — verify prettier works with the ignore file
- [ ] Verify `.husky/pre-commit` ends with a newline


jshaughn (approved):
Good fix. Do we need more things under .prettierignore? I guess not given the applicable file extensions.

PR #793, opened 2026-08-03
fix: load monaco-editor locally instead of from cdn.jsdelivr.net

### LINKED ISSUE #790: [OSSMC] Monaco editor loads worker scripts from external CDN instead of local bundle
## Problem

The `@monaco-editor/react` package (v4.7.0) defaults to fetching Monaco's web worker scripts from `cdn.jsdelivr.net` at runtime, even though `monaco-editor` (v0.55.1) is already installed as a local dependency. This causes:

1. **Air-gapped / disconnected OpenShift environments**: Users behind firewalls or in disconnected clusters cannot use the YAML editor at all since it requires live HTTP access to `cdn.jsdelivr.net`. This is especially relevant for OSSMC which frequently runs in restricted enterprise environments.

2. **Potential CI test flakes**: Any Cypress tests that navigate to pages with Monaco editors could fail when `cdn.jsdelivr.net` is slow or unreachable from CI runners, producing uncaught `importScripts` errors.

3. **Unnecessary network dependency**: Every page load with an editor component makes external HTTP requests for a library that is already bundled in `node_modules`.

## Example error (from Kiali CI with same dependencies)

```
(uncaught exception) Error: UncaughtNetworkError: Failed to execute 'importScripts'
on 'WorkerGlobalScope': The script at
'https://cdn.jsdelivr.net/npm/monaco-editor@0.55.1/min/vs/assets/editor.worker-Be8ye1pW.js'
failed to load.
```

## Suggested Fix

Add a webpack `resolve.alias` in `plugin/webpack.config.ts` to redirect `'monaco-editor'` imports to the worker-free `editor.api` entry point, and configure `@monaco-editor/react`'s loader to use the local instance:

**webpack.config.ts:**
```typescript
resolve: {
  alias: {
    'monaco-editor$': path.resolve(__dirname, 'node_modules/monaco-editor/esm/vs/editor/editor.api.js')
  },
  // ... existing config
}
```

**Entry point (or a shared setup file):**
```typescript
import * as monaco from 'monaco-editor';
import { loader } from '@monaco-editor/react';

loader.config({ monaco });
```

The `editor.api` entry provides all exports OSSMC needs (`editor`, `MarkerSeverity`, `Selection`, `languages`) without bundling web workers. Language services run on the main thread, which is fine for Kiali/OSSMC's small YAML/JSON config editing.

This approach was validated in kiali/kiali#10131 where it also reduced the production bundle by ~50%.

## Related

- kiali/kiali#10130 — Same issue in Kiali standalone
- kiali/kiali#10131 — Fix PR for Kiali standalone

## Affected files

Files that import from `monaco-editor` or `@monaco-editor/react`:
- `plugin/src/kiali/components/IstioConfigPreview/EditorPreview.tsx`
- `plugin/src/kiali/components/Envoy/EnvoyDetails.tsx`
- `plugin/src/kiali/pages/Mesh/target/TargetPanelEditor.tsx`
- `plugin/src/kiali/components/Mesh/CheckerTracingConfig.tsx`
- `plugin/src/kiali/types/EditorValidations.ts`

ferhoyos:
## Summary

- Alias `monaco-editor` to the `editor.api` entry point in `webpack.config.ts`, so `@monaco-editor/react` loads the locally bundled `monaco-editor` instead of fetching it from `cdn.jsdelivr.net` at runtime.
- Add `src/openshift/utils/MonacoSetup.ts`, which configures `@monaco-editor/react`'s loader via `loader.config({ monaco })`, explicitly imports the YAML basic-language contribution (excluded by `editor.api`), and sets a self-contained `MonacoEnvironment.getWorker` so OSSMC's Monaco instance does not inherit OpenShift Console's global worker configuration.
- Import `MonacoSetup` from `KialiContainer`, which every OSSMC page renders through, ensuring the loader and worker setup run before any page can mount a Monaco editor.

## Problem

`@monaco-editor/react` (v4.7.0) defaults to fetching Monaco from `cdn.jsdelivr.net` at runtime, even though `monaco-editor` (v0.55.1) is already installed as a local dependency. This causes:

1. **Air-gapped / disconnected OpenShift environments**: Users behind firewalls or in disconnected clusters can't use the YAML editor at all since it requires live HTTP access to `cdn.jsdelivr.net`.
2. **Potential CI test flakes**: Cypress tests that navigate to pages with Monaco editors could fail when `cdn.jsdelivr.net` is slow or unreachable.
3. **Unnecessary network dependency**: Every page load with an editor component makes external HTTP requests for a library that's already bundled in `node_modules`.

Additionally, once Monaco is loaded locally, OSSMC still shares the browser realm with OpenShift Console. Console bundles its own, separately versioned `monaco-editor` and (via `monaco-editor-webpack-plugin`'s `globalAPI: true`) sets a page-wide `window.MonacoEnvironment.getWorkerUrl`. Without our own `getWorker`, OSSMC editor instances would inherit Console's worker URL and speak an incompatible RPC protocol to a differently versioned worker script.

This mirrors [kiali/kiali#10131](https://github.com/kiali/kiali/pull/10131), which fixed the CDN issue in the Kiali standalone frontend (`rsbuild.config.ts` + `index.tsx`). OSSMC needs its own equivalent fix since it uses a different build (webpack, not Rsbuild) and has no traditional entry point — pages are lazily loaded via `console-extensions.ts` `$codeRef`s, so `KialiContainer` (rendered by every OSSMC page) is used as the initialization point instead. The Console `MonacoEnvironment` collision is OSSMC-specific and has no standalone-Kiali counterpart.

## Test plan

- [x] `yarn build` completes successfully with the alias in place
- [x] `yarn lint` passes with no new errors
- [x] `tsc --noEmit` passes with no errors
- [x] `yarn test src/openshift` passes
- [x] Verified the production bundle no longer emits the full `editor.main` worker bundle graph for the aliased `monaco-editor` import
- [x] Local OSSMC: YAML editor opens without CDN requests and without inheriting Console's Monaco worker

Fixes #790

### DIFF
--- a/plugin/src/openshift/components/KialiContainer.tsx
+++ b/plugin/src/openshift/components/KialiContainer.tsx
@@ -1,17 +1,18 @@
 import * as React from 'react';
 import { Provider } from 'react-redux';
+import { classes } from 'typestyle';
+
 import { store } from 'store/ConfigStore';
+import { globalStyle as kialiStyle } from 'styles/GlobalStyle';
+import kialiCSSVariables from 'styles/variables.module.scss';
+
 import { KialiController } from '../components/KialiController';
 import { NotificationAlerts } from '../components/NotificationAlerts';
-import { globalStyle as kialiStyle } from 'styles/GlobalStyle';
 import { globalStyle as ossmcStyle } from '../styles/GlobalStyle';
-import kialiCSSVariables from 'styles/variables.module.scss';
 import ossmcCSSVariables from '../styles/variables.module.scss';
 
-// Load the pf-icons
 import '@patternfly/patternfly/patternfly-base.css';
-
-import { classes } from 'typestyle';
+// Load the pf-icons
 
 // Configure @monaco-editor/react to use the locally bundled monaco-editor before any page
 // can render an editor. See MonacoSetup.ts for details.
--- a/plugin/src/openshift/components/KialiContainer.tsx
+++ b/plugin/src/openshift/components/KialiContainer.tsx
@@ -13,6 +13,10 @@ import '@patternfly/patternfly/patternfly-base.css';
 
 import { classes } from 'typestyle';
 
+// Configure @monaco-editor/react to use the locally bundled monaco-editor before any page
+// can render an editor. See MonacoSetup.ts for details.
+import '../utils/MonacoSetup';
+
 interface Props {
   children: React.ReactNode;
   className?: string;
--- a/plugin/src/openshift/components/KialiContainer.tsx
+++ b/plugin/src/openshift/components/KialiContainer.tsx
@@ -11,8 +11,8 @@ import { NotificationAlerts } from '../components/NotificationAlerts';
 import { globalStyle as ossmcStyle } from '../styles/GlobalStyle';
 import ossmcCSSVariables from '../styles/variables.module.scss';
 
-import '@patternfly/patternfly/patternfly-base.css';
 // Load the pf-icons
+import '@patternfly/patternfly/patternfly-base.css';
 
 // Configure @monaco-editor/react to use the locally bundled monaco-editor before any page
 // can render an editor. See MonacoSetup.ts for details.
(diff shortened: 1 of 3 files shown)


jshaughn (approved):
Yup, thanks @ferhoyos !

PR #795, opened 2026-08-03
Fix several UI/navigation bugs found in Kiali frontend

### LINKED ISSUE #10151: Several small UI/navigation bugs in the Kiali frontend
While working on other changes, I found several small, unrelated bugs in the vendored Kiali frontend code (`plugin/src/kiali`). Filing this issue to track them; a PR with fixes is here: https://github.com/kiali/openshift-servicemesh-plugin/pull/795

## Bugs found

1. **`MiniGraphCard`: broken query string when returning from a node's mini-graph**
   `router.navigate` builds the target URL with `` `${location.getPathname()}?&{urlParams.toString()}` `` — the `$` is missing before `{urlParams...}`, so the literal string `?&{urlParams.toString()}` is appended to the URL instead of the actual query string.
   **Fix:** correct the template literal to `` `${location.getPathname()}?${urlParams.toString()}` `` so the URL params are properly interpolated.

2. **`SimpleTable`: cell objects render as `[object Object]`**
   When a row's `cells` array contains an object shaped like `{ title: ReactNode }` (as opposed to a plain `ReactNode`), the cell renders the object directly, which stringifies to `[object Object]` instead of showing the intended content.
   **Fix:** detect cells shaped as `{ title }` and render `cell.title` instead of the raw cell value.

3. **`DataPlaneStats`: footer link disappears during refresh**
   The card footer (with the "view data planes" link) is hidden any time `isCardLoading` is true, which includes brief health-only refreshes. This can cause the link to disappear out from under a user mid-click.
   **Fix:** introduce a `showFooter` flag that only hides the footer when namespaces are still loading or the card is in an error state, keeping it visible during health-only refreshes.

4. **Notification drawer collapse causes horizontal scroll/clipping**
   When the notification drawer is collapsed, PatternFly still reserves its `FlexBasis` width. Certain focus/automation interactions can set `scrollLeft` on `.pf-v6-c-drawer__main`, shifting the whole page horizontally under the (invisible) reserved panel space and clipping toolbar controls and table headers.
   **Fix:** add CSS to fully collapse the panel's width/flex-basis when not expanded, and add a `useEffect` in `Navigation` that resets `scrollLeft` back to `0` on the drawer's main container whenever the drawer is collapsed or the route changes.

5. **`GraphHelpTour` / `MeshHelpTour`: Shortcuts tour stop renders nothing useful**
   `htmlDescription` is set to `GraphShortcuts` / `MeshShortcuts` — a reference to the function itself — instead of calling it. Since `htmlDescription` expects rendered content (`React.ReactNode`), this renders a stray function reference instead of the shortcuts list.
   **Fix:** call the functions (`GraphShortcuts()` / `MeshShortcuts()`) so their actual rendered output is used.

6. **`GraphToolbar`: node-graph toolbar briefly renders as the full graph toolbar**
   `isNodeGraph` is derived purely from `this.props.node`, which is set by `GraphPage.componentDidMount` via Redux. On the very first paint of a node-graph route, Redux can lag one frame behind the URL, so the toolbar briefly renders as if it were the full graph (missing the "back to full graph" control, wrong graph-type behavior).
   **Fix:** compute `isNodeGraph` by also checking the current URL path (`location.getPathname().includes('/graph/node')`) as a fallback, so the correct toolbar renders even before Redux state catches up.

7. **`styleGroup`: possible crash reading bounds of a group node**
   The component reads `element.getBounds()` into a ref and then unconditionally accesses `.x`/`.y`/`.width`/`.height` on it when rendering the focus overlay rectangle, without checking whether the bounds are actually available yet.
   **Fix:** add a null check on the bounds before rendering the focus overlay `<rect>`, so no overlay is rendered (instead of crashing) when bounds aren't available.

## Notes

These were all discovered incidentally while working on unrelated changes and have no connection to each other beyond living in the same general area of the frontend.

Generated-by: Cursor

jmazzitelli:
Fixes a handful of small, unrelated bugs found while working on other changes elsewhere in the codebase:

- MiniGraphCard: fixed a query-string typo that produced a literal `?&{urlParams...}` in the URL instead of interpolating the params.
- SimpleTable: cells shaped as `{ title }` objects now render their title instead of `[object Object]`.
- DataPlaneStats: the "view data planes" footer link now stays visible during health-only refreshes instead of disappearing mid-click.
- Navigation/GlobalStyle: the collapsed notification drawer no longer steals horizontal scroll and clips toolbar/table content.
- GraphHelpTour/MeshHelpTour: the Shortcuts tour stop now renders the shortcuts content instead of a stray function reference.
- GraphToolbar: the node-graph toolbar no longer briefly renders as the full graph on first paint, by preferring the URL over Redux state.
- styleGroup: added a null check to avoid a possible crash when a group node's bounds aren't yet available.

fixes: https://github.com/kiali/kiali/issues/10151

Generated-by: Cursor

### DIFF
--- a/plugin/src/kiali/components/Nav/Navigation.tsx
+++ b/plugin/src/kiali/components/Nav/Navigation.tsx
@@ -72,6 +72,24 @@ export const NavigationComponent: React.FC<NavigationProps> = (props: Navigation
     document.title = pageTitle;
   }, []);
 
+  // Keep page content aligned when the notification drawer is collapsed. Focus or
+  // automation can scroll drawer__main horizontally into the reserved panel slot.
+  React.useEffect(() => {
+    if (props.showNotificationCenter) {
+      return;
+    }
+
+    const resetDrawerScroll = (): void => {
+      document.querySelectorAll('.pf-v6-c-drawer__main').forEach(el => {
+        (el as HTMLElement).scrollLeft = 0;
+      });
+    };
+
+    resetDrawerScroll();
+    const id = requestAnimationFrame(resetDrawerScroll);
+    return () => cancelAnimationFrame(id);
+  }, [props.showNotificationCenter, pathname]);
+
   const isGraph = (): boolean => {
     return pathname.startsWith('/graph') || pathname.startsWith('/mesh');
   };
--- a/plugin/src/kiali/components/Table/SimpleTable.tsx
+++ b/plugin/src/kiali/components/Table/SimpleTable.tsx
@@ -119,11 +119,22 @@ export const SimpleTable: React.FC<SimpleTableProps> = (props: SimpleTableProps)
         {props.rows.length > 0 ? (
           props.rows.map((row, rowIndex) => (
             <Tr key={row.key ?? `row_${rowIndex}`} className={row.className}>
-              {row.cells?.map((cell: React.ReactNode, colIndex: number) => (
-                <Td key={`cell_${rowIndex}_${colIndex}`} dataLabel={props.columns[colIndex].title} className={tdStyle}>
-                  {cell}
-                </Td>
-              ))}
+              {row.cells?.map((cell, colIndex: number) => {
+                const cellContent =
+                  typeof cell === 'object' && cell !== null && 'title' in cell
+                    ? (cell as { title: React.ReactNode }).title
+                    : (cell as React.ReactNode);
+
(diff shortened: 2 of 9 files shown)


jmazzitelli:
wrong repo - these are kiali server changes - need to be made over there.

PR #796, opened 2026-08-03
OSSM-ACM - Refresh core dependencies while staying on React 17 [stack 1 of 3]

jmazzitelli:
OpenShift console plugins are not supported on React 18 prior to OpenShift 4.22, and this branch needs to support 4.20, so react/ react-dom stay pinned to 17.x. Refreshes react-router(-dom), redux(-thunk), react-redux, typesafe-actions, and i18next to the latest versions compatible with React 17, and bumps the OpenShift console dynamic-plugin-sdk. Adds resolutions for @types/react(-dom), redux, and react-i18next so the newly-added dependency tree doesn't drag in a second, incompatible copy of any of them. Also updates the ConsolePlugin manifest API version and CI/build tooling (Makefile.test.mk, CI install hardening).


jmazzitelli:
never mind - stacked PRs don't work when your branches are in our fork.

* https://docs.github.com/en/pull-requests/how-tos/create-pull-requests/creating-stacked-pull-requests
  * `Stacked pull requests require all branches to be in the same repository. Cross-fork stacks are not supported.`

* https://github.com/github/gh-stack/issues/46

